PR #17, opened 2016-11-10
Carl/58 connect frontend to backend login service

MasterCarl:


Langleu:
not working if you are behind a strict firewall, or well everything is blocked :) not your fault!
Would be nice if the ip in the feather.js would be saved in a config for production/default/test

pinussilvestrus:
fix: http://stackoverflow.com/questions/11916999/socket-io-only-works-locally

PR #39, opened 2016-11-22
replaced cookies with local storage + client part of activity based permission system

NicoKnoll:


jprberlin:
Has this any browser min requirements?

NicoKnoll:
@jprberlin nope, we use babel polyfill to make it work with every modern browser (not netscape and stuff)

NicoKnoll:
http://caniuse.com/namevalue-storage/embed/description

PR #48, opened 2016-12-01
create core-module and helper for notification popups

pinussilvestrus:
All Notifications (Info, warning or error-messages) can no be displayed by the notification-helper

image


src/modules/core/helpers/notification.js
```diff
@@ -0,0 +1,24 @@
...
+
+	showError(message) {
+		Alert.error(message, {
+			position: 'top',
+			effect: 'jelly',
+			beep: false,
+			timeout: 'none'
+		});
+	}
+	showNormal(message) {
```
nicoknoll: showInfo

src/modules/core/helpers/index.js
```diff
@@ -2,10 +2,12 @@ import App from './app';
 import Config from './config';
```
nicoknoll: Move package to core instead of static :)

PR #65, opened 2017-01-12
Add new component with the system selector/button + actions

MasterCarl:
The elements are now hidden by default, the 'local' system is selected in this case


src/modules/login/actions/login.js
```diff
@@ -3,6 +3,21 @@ import {browserHistory} from 'react-router';
 import { Server, Notification } from '../../core/helpers';
 
 const authService = Server.service('/auth');
+const schoolService = Server.service('/schools');
+const systemService = Server.service('/systems');
+
+String.prototype.capitalizeFirstLetter = function () {
```
nicoknoll: Whoa maybe don't extend the String class here but just make it either make it a normal function or put it somewhere so it's available for everything.

src/modules/login/actions/login.js
```diff
@@ -3,6 +3,21 @@ import {browserHistory} from 'react-router';
...
+	return this.substr(0, 1).toUpperCase() + this.substr(1);
+};
+
+Array.prototype.indexBy = function (property) {
```
nicoknoll: Same

src/modules/login/components/login_form.jsx
```diff
@@ -1,12 +1,16 @@
...
 
 	constructor(props) {
 		super(props);
 
 		this.state = {
 			email: '',
-			password: ''
-		}
+			password: '',
+			systemId: '0000d186816abba584714c92'	// the id of the 'local' system.
```
nicoknoll: We should discuss with @Langleu how it should work for the local system as it isn't related to any school

src/modules/login/components/login_form.jsx
```diff
@@ -20,72 +24,25 @@ class LoginForm extends React.Component {
 		this.props.actions.login.bind(this)({
 			email: this.state.email,
 			password: this.state.password,
-			school: this.state.school || undefined,
```
nicoknoll: why remove school?

MasterCarl: I removed it since we didn't use the parameter in the login action. However, we should actually pass it to the account service since there's a field `schoolId` in our model. I'll put it back :)

PR #66, opened 2017-01-14
191 signup flow schulen und lehrer

nicoknoll:
It got a little bit more complex than expected. So this PR includes:

* Updated all NPM modules (especially `feathers-authentication` with breaking changes)

* Implemented `Formsy` and `React-Select` in a helper component
* Implemented Signup (before login) and Setup (after login) pages
* Implemented signup flow ("create account, then create user, ...")
* Implemented `displayName` for users (shown in topbar)
* Implemented `useScroll` for Router for native browser feeling (jump to top of page if link is clicked)

* Improved style and elements of sidebar
* Improved login flow

* Refactored AdminSections to be reusable
* Refactored connection between `accounts` and `users`


src/modules/administration/containers/administration.js
```diff
@@ -14,6 +14,14 @@ const roleService = Server.service('/roles');
 import component from '../components/administration';
 import actions from '../actions/administration';
 
+const pluckArrayToObject = (array, key) => {
```
MasterCarl: I'd call this something like `indexArrayByKey`

PR #67, opened 2017-01-18
newtool form now uses formsy

pinussilvestrus:


src/modules/tools/components/newToolForm.jsx
```diff
@@ -94,47 +89,61 @@ class NewToolForm extends React.Component {
...
-				</label> <br></br>
+			<Form className="new_tool_form">
+				<Select
+					label="Kurs"
+					name="courseId"
+					type="text"
+					options={this.getCourseOptions()}
+					value={this.state.tool.courseId}>
+				</Select>
+			 	<br></br>
```
nicoknoll: Why <br>? Just add layout="vertical" and it will wrap it in bootstrap rows

nicoknoll: Why <br>? Just add layout="vertical" and it will wrap it in bootstrap rows

PR #88, opened 2017-02-10
Use Calendar service in client

nicoknoll:


nicoknoll:
[Use Calendar Service in Client](https://trello.com/c/xtqqzkRr/302-use-calendar-service-in-client)

src/modules/timetable/components/calendar.jsx
```diff
@@ -32,6 +32,7 @@ class Calendar extends React.Component {
 				left: 'title',
 				right: 'month,agendaWeek,agendaDay prev,today,next'
 			},
+			height: 750,
```
MasterCarl: I have a feeling that should go in the SCSS

src/modules/timetable/containers/calendar.js
```diff
@@ -1,20 +1,52 @@
...
 	if(!['month','agendaWeek','agendaDay'].includes(view)) {
 		view = 'month';
 	}
 
-	let componentData = {
-		events: [],
-		view
-	};
+	let error;
+	const getEvents = new Promise((resolve, reject) => {
```
MasterCarl: Wrapping this in a new Promise is unnecessary, you could just write `const getEvents = calendarService.find().catch(err => {return []});`

PR #93, opened 2017-02-28
Added CSS Breakpoints

adrianjost:
Added new CSS Breakpoint to view text on displays larger than `media-breakpoint-up(lg)`

### DIFF
--- a/src/modules/base/components/sidebar.jsx
+++ b/src/modules/base/components/sidebar.jsx
@@ -17,7 +17,7 @@ class Sidebar extends React.Component {
 	render() {
 		const sidebarLinks = [{
 			name: 'Dashboard',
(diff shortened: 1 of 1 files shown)


adrianjost:
[Have a version of Left Hand Nav with text if screen size is big enough..](https://trello.com/c/AeWuRsWR/309-have-a-version-of-left-hand-nav-with-text-if-screen-size-is-big-enough)

pinussilvestrus:
can this be closed? because https://github.com/schulcloud/schulcloud-client/pull/95 seems to also implement it, while checking out this branch locally, the larger text isn't displayed...

adrianjost:
yes, this can be closed. You are right, #95 also implement this.

PR #95, opened 2017-03-07
306 left nav should keep active states when navigating to subitems like class

adrianjost:
Changed 
`activeClassName="active" `
to 
`location.pathname.startsWith(link.to) && 'active'`
so there should be always an matching state in sidebar.


adrianjost:
[Left nav should keep active states when navigating to subitems like class.](https://trello.com/c/d3BQAeeC/306-left-nav-should-keep-active-states-when-navigating-to-subitems-like-class)

PR #103, opened 2017-03-14
Add default MIME type value 🎫

MasterCarl:


pinussilvestrus:
[permit file upload if file.type is undefined](https://trello.com/c/XseJLfRU/255-permit-file-upload-if-file-type-is-undefined)

PR #105, opened 2017-03-15
Sticky Footer

adrianjost:
Footer is now always sticky to the bottom but never overlaps content


adrianjost:
[Sticky Footer](https://trello.com/c/BQtI3h0a/339-sticky-footer)

adrianjost:
[Sticky Footer](https://trello.com/c/BQtI3h0a/339-sticky-footer)

PR #114, opened 2017-03-20
html5 password validation

Langleu:


pinussilvestrus:
[Usernames should be lowercased, form should disable automatic capitalization](https://trello.com/c/8awzqeAL/300-usernames-should-be-lowercased-form-should-disable-automatic-capitalization)

PR #120, opened 2017-03-27
display current user role

pinussilvestrus:


pinussilvestrus:
[Show Role of current user in the upper right corner](https://trello.com/c/c6jPoNUh/352-show-role-of-current-user-in-the-upper-right-corner)

PR #123, opened 2017-03-27
ressurect header/old footer 😇

Langleu:


Langleu:
[resurrect footer(error/login page)/header](https://trello.com/c/IMDVeYG0/356-resurrect-footer-error-login-page-header)

PR #130, opened 2017-03-29
fix lti run

pinussilvestrus:


pinussilvestrus:
[LTI Run doesn't work (timeout)](https://trello.com/c/i0laW6ZK/375-lti-run-doesn-t-work-timeout)

PR #135, opened 2017-03-31
sidebar highlight does work without ending /

adrianjost:


adrianjost:
[Sidebar highlight does not work without ending /](https://trello.com/c/aYZGwAKQ/365-sidebar-highlight-does-not-work-without-ending)

PR #140, opened 2017-04-04
increase timeout

pinussilvestrus:


pinussilvestrus:
[frontend login test randomly fail](https://trello.com/c/WIiuN9fN/390-frontend-login-test-randomly-fail)

PR #161, opened 2017-04-18
add alias to systems

pinussilvestrus:


pinussilvestrus:
[Give systems an alias](https://trello.com/c/BfK4j6qs/416-give-systems-an-alias)

PR #165, opened 2017-04-19
Add Docker for local development

kremer-io:
Could someone test this under Windows?


Langleu:
what about gulp or the fact, that nodemon does not notice all files changes 100% ?

kremer-io:
@Langleu gulp is used for building assets (JS, CSS, Images). nodemon is used to restart NodeJS if any file did change, so that development is easier. Right? In my experience nodemon is working fine, do you have any examples regarding noticing file changes?

@Langleu But we should consider building assets inside the container. As far as I know, there is additional work needed for running multiple tasks in one container. E.g. supervisord. Another solution could be an extra container for gulp watch.

Langleu:
@kremer-io First of all you're right, but...
Nodemon often gets stuck noticing .hbs file changes, meaning you will have to restart the docker container all the time and our client is pretty much .hbs and .js (gulp).

So you have to run gulp from outside the container? That means the user is left again installing node, npm and gulp --> docker is kinda obsolete in that case, except as you noticed yourself already you figure a way out to run multiple tasks in one container through maybe Tmux, screen or supervisord as you said yourself already.

kremer-io:
@Langleu The user should run `docker-compose up`. This should include building assets and restarting the server when needed. I'll setup asset management inside the container. Do you know any workaround for the nodemon handlebars problem?

Langleu:
@kremer-io didn't figure out a workaround yet for nodemon and handlebars.
My good guess would be to leave the client alone and rather write an easy tutorial for the client...
the work you will put in for figuring out all those workarounds (gulp, nodemon, ...) is not worth the time.

The dockerized server, that you wrote,  on the other hand is quite good and will work with the client aslong as the port is bound to localhost.

kremer-io:
@Langleu Ok, but we should still dockerize the client, even without nodemon handlebars support. I'll limit my efforts here. There will be users who just needs a GUI for there stuff to test - e.g. content search. They can use the Client, even without nodemon and gulp.

Langleu:
@kremer-io I only see a very limited use case for that but ofc go on.
Please then do the same for gulp as you did for the node_modules so the user can at least see the GUI

kremer-io:
@Langleu atm, gulp is run once - but could be switched to `gulp watch`

pinussilvestrus:
@kremer-io @Langleu thanks to #191 gulp watch does not crash on js-errors anymore, so you can use it.

pinussilvestrus:
Habs auch mal auf Windows getestet und bekomme folgenden Error (ähnlich wie im Server)
```
ERROR: for schulcloud-client  Cannot create container for service schulcloud-client: invalid bind mount spec "D:\\GitHub\\schulcloud\\schulcloud-client:/app:rw": invalid volume specification: 'D:\GitHub\schulcloud\schulcloud-client:/app:rw'

ERROR: for schulcloud-client-gulp  Cannot create container for service schulcloud-client-gulp: invalid bind mount spec "D:\\GitHub\\schulcloud\\schulcloud-client:/app:rw": invalid volume specification: 'D:\GitHub\schulcloud\schulcloud-client:/app:rw'
ERROR: Encountered errors while bringing up the project.
```

jprberlin:
wie ist denn hier der Status?

PR #178, opened 2017-04-21
Update login.scss

pinussilvestrus:


pinussilvestrus:
According to @kremer-io 's issue that these images are not correctly referrenced in a docker container.

PR #190, opened 2017-04-25
send email for students

Langleu:


Langleu:
[Send registration emails when new user are created](https://trello.com/c/Ws9z9rhh/425-send-registration-emails-when-new-user-are-created)

PR #206, opened 2017-05-05
Refactor file access

MasterCarl:


pinussilvestrus:
[Merge filename and storageContext into "path"](https://trello.com/c/IGV2d2X8/467-merge-filename-and-storagecontext-into-path)

PR #208, opened 2017-05-07
468 if dates changed on course edit also delete add calendar entries events

pinussilvestrus:


pinussilvestrus:
[If dates changed on course edit also delete/add calendar entries (events)](https://trello.com/c/bmQ3yd2s/468-if-dates-changed-on-course-edit-also-delete-add-calendar-entries-events)

PR #209, opened 2017-05-08
fix event-course link

pinussilvestrus:


pinussilvestrus:
[On course-event-click: course is undefined](https://trello.com/c/Xsrq9jF1/487-on-course-event-click-course-is-undefined)

PR #228, opened 2017-05-16
Implemented jQuery datepicker

nicoknoll:


pinussilvestrus:
[implement fallback for dateselection in firefox and safari](https://trello.com/c/lQwjxjGx/413-implement-fallback-for-dateselection-in-firefox-and-safari)

PR #236, opened 2017-05-18
add room to course times

pinussilvestrus:


pinussilvestrus:
[Add Room to course-model](https://trello.com/c/SCATXb6H/521-add-room-to-course-model)

PR #247, opened 2017-05-23
player js, video skipping, refactoring

davidkreidler:


pinussilvestrus:
[(Video-) file streaming](https://trello.com/c/iuYpmkjR/530-video-file-streaming)

static/scripts/files.js
```diff
@@ -225,20 +225,30 @@ $moveModal.modal('hide');
     });
 
 });
+    function videoClick(e) {
+        e.stopPropagation();
+        e.preventDefault();
+    }
 
     function fileViewer(filetype, file) {
+        document.getElementById('my-video').style.display = "none";
```
pinussilvestrus: Works fine! 👍  Just a little code 'request-for-change', could you use jquery for handeling html-elements? Because we do so in nearly all other js-files. So instead of

```js
document.getElementById('my-video').style.display = "none";
```

you can use

```js
$('#my-video').css("display","");
```

And so on, on many other places of your code.

PR #249, opened 2017-05-24
fix for bootstrap and tether

pinussilvestrus:


pinussilvestrus:
That does not fix it 100%, because when you go directly on a page via url (not via navigation), it still breaks.

image

For the weekly-challenge its ok, but it would be nice if its work perfectly later on 😄

PR #252, opened 2017-05-29
also delete account if student/teacher is deleted

pinussilvestrus:


pinussilvestrus:
[When deleting a User in Admin-Area, the Account will not be deleted](https://trello.com/c/Y60GW0f8/536-when-deleting-a-user-in-admin-area-the-account-will-not-be-deleted)

PR #275, opened 2017-06-01
Flexbox

nicoknoll:


Langleu:
[Use flexbox design with bootstrap](https://trello.com/c/6CfVeiKa/558-use-flexbox-design-with-bootstrap)

PR #288, opened 2017-06-08
do everything in utc

Langleu:


Langleu:
everything works as intended regarding time zones and saving them in utc.
New approach would be to show everything in utc and do as the local timezone is the utc timezone, so we get rid of summer/winter time and all that stuff.

pinussilvestrus:
[Zeit bei Kursterminen ist um 2 Stunden zeitversetzt.](https://trello.com/c/KQet9xHC/571-zeit-bei-kursterminen-ist-um-2-stunden-zeitversetzt)

Langleu:
thanks to @nicoknoll and @MasterCarl for their input

PR #293, opened 2017-06-09
fix paddings

kremer-io:


jprberlin:
We will do further tweaks in this branch

PR #295, opened 2017-06-09
substition teacher v1

Langleu:
For further versions think of a better version, but due to limited time this should be pleasing enough.
Maybe in further versions a button to assign a teacher for just one lesson as substitution?


pinussilvestrus:
[Vertretungslehrer hinzufügen](https://trello.com/c/37fF6mqk/585-vertretungslehrer-hinzuf%C3%BCgen)

PR #305, opened 2017-06-20
fix dropzone

pinussilvestrus:


pinussilvestrus:
[Click on dropzone-label does not open file selector (just click around)](https://trello.com/c/fMgcWJM0/570-click-on-dropzone-label-does-not-open-file-selector-just-click-around)

PR #321, opened 2017-07-04
619 themen export import teilen

pinussilvestrus:
DEPENDS ON: https://github.com/schul-cloud/schulcloud-server/pull/187


pinussilvestrus:
[Themen Export/Import/Teilen](https://trello.com/c/a1XxDsXl/619-themen-export-import-teilen)

Langleu:
you could generate one shareToken for each topic instead of generating each time a new token.

PR #328, opened 2017-07-13
629 -unique editor design with slightly changed layout

adrianjost:
require changes from (color picker):
https://github.com/schul-cloud/schulcloud-client/pull/319/

editor-config


adrianjost:
[Gleicher Editor in Themen und Aufgaben](https://trello.com/c/hjFIBt9l/629-gleicher-editor-in-themen-und-aufgaben)

PR #334, opened 2017-07-24
460 proxy database file storage

pinussilvestrus:
depends on: https://github.com/schul-cloud/schulcloud-server/pull/193


pinussilvestrus:
[Proxy Database to store the storage context for each file so we don't have to pass it in the url](https://trello.com/c/u2pslIun/460-proxy-database-to-store-the-storage-context-for-each-file-so-we-dont-have-to-pass-it-in-the-url)

PR #347, opened 2017-08-07
select your own gender

Langleu:


Langleu:
[Add "gender" to User-Model (Settings, Registration)](https://trello.com/c/V9vOaTmg/658-add-gender-to-user-model-settings-registration)

PR #348, opened 2017-08-07
change wording according to role

Langleu:


Langleu:
[change wording of teacher-admin area](https://trello.com/c/HRnHfGKo/659-change-wording-of-teacher-admin-area)

PR #349, opened 2017-08-07
save most scripts

Langleu:
Saved most scripts and css, except for MathJax, as this contains around 66k documents (dependencies for specific functionality).


Langleu:
[Deliver assets from our servers to avoid blacklists on local firewalls](https://trello.com/c/kF8KT04v/644-deliver-assets-from-our-servers-to-avoid-blacklists-on-local-firewalls)

PR #363, opened 2017-08-18
Piwik custom tracking

Langleu:


static/scripts/piwik/piwik.js
```diff
@@ -1,6 +1,15 @@
 var _paq = _paq || [];
-_paq.push(['trackPageView']);
-_paq.push(['enableLinkTracking']);
+$.get('/account/user', (res) => {
+    if (typeof (res || {}).firstName === 'string') {
```
jprberlin: is this how to detect if the user is logged in? If so add an comment

static/scripts/piwik/piwik.js
```diff
@@ -1,6 +1,15 @@
...
+        _paq.push(['setCustomDimension', customDimensionId = 2, customDimensionValue = res.gender]);
+        _paq.push(['setCustomDimension', customDimensionId = 4, customDimensionValue = res.roles[0].name]);
+    }
+
+    _paq.push(['trackPageView']);
+    _paq.push(['enableLinkTracking']);
+});
+
 (function () {
     var u = "//open.hpi.de/piwik/";
```
jprberlin: should this go to some global config?

Langleu: no

static/scripts/piwik/piwik.js
Langleu: will change it a bit

PR #414, opened 2017-09-19
739-automatically-change-invalid-values-to-100-or-0-with-the-new-percentage-system

adrianjost:


adrianjost:
[automatically change invalid values to 100 or 0 with the new percentage system](https://trello.com/c/jTTFb3e0/739-automatically-change-invalid-values-to-100-or-0-with-the-new-percentage-system)

PR #430, opened 2017-09-25
introduce NOTIFICATION_SERVICE_ENABLED

Langleu:


adrianjost:
[introduce notificaiton env for client, less errors](https://trello.com/c/r7fGFFua/763-introduce-notificaiton-env-for-client-less-errors)

PR #455, opened 2017-10-16
add sorting for tables and classes for students

Langleu:
everything is sortable but classes for students/teachers as those are not anyhow related to the object


Langleu:
[Schülerverwaltung](https://trello.com/c/Xn0yMeQY/778-sch%C3%BClerverwaltung)

PR #460, opened 2017-10-26
change error message and change hint

Langleu:


Langleu:
[Change error message of password recovery in case a non existing username is entered.](https://trello.com/c/0X6qAxsO/790-change-error-message-of-password-recovery-in-case-a-non-existing-username-is-entered)

PR #490, opened 2017-11-14
Added anonymized google analytics tracking

soldag:
Added middleware which tracks page views and sends hits to Google Analytics via Measurement Protocol.


helpers/googleAnalytics.js
```diff
@@ -0,0 +1,116 @@
...
+  request({
+    method: 'POST',
+    uri: BATCH_HITS_ENDPOINT,
+    encoding: 'utf-8',
+    body,
+  }).catch(err => console.error(`Error while emitting Google Analytics hits (${err.message})`));
+};
+
+const anonymizeIp = (ip) => {
+  ip = ipaddr.parse(ip);
```
jprberlin: can parsing fail under some circumstances?

helpers/googleAnalytics.js
```diff
@@ -0,0 +1,116 @@
...
+      emit(hit);
+    });
+  }
+
+  next();
+};
+
+// Flush hit queue on exit
+process.on('exit', () => {
+  console.log('Sending queued Google Analytics hits...');
```
jprberlin: do we really want logging here each time?

helpers/googleAnalytics.js
soldag: Only if the request IP is not valid, but I can also add error handling just in case.

helpers/googleAnalytics.js
soldag: This handler is only executed when the server is shutdown and is used to sent remaining queued hits to GA. I thought it would be nice to print this message so nobody gets confused of the delay when shutting down the server.

jprberlin: ok, than its fine for me

jprberlin:
@Langleu After review this could be released and the ENV Var should be set to: UA-109651848-1

PR #497, opened 2017-11-24
add password check for account creation

Langleu:
now also exists for account/full, meaning either invited through a person or the general invitation link


Langleu:
[Doppelte Password Eingabe beim Anlegen von Accounts](https://trello.com/c/yQ9I7PUX/837-doppelte-password-eingabe-beim-anlegen-von-accounts)

PR #499, opened 2017-11-28
Fix search pagination, use $limit & $skip

kremer-io:
Only works with schulcloud-content branch [fix-search-pagination](https://github.com/schul-cloud/schulcloud-content/tree/fix-search-pagination)


jprberlin:
then please open up the related PR on the content service as well

PR #500, opened 2017-11-28
Add referer and geoip lookup to google analytics tracking. 

soldag:
This PR adds a GeoIP lookup to retrieve the country of an user from its IP before it gets anonymized. The retrieved country and the Referer header, if set, is added to the Google Analytics hit. 
Additionally, an issue with the scheduling of flushing the hit queue was fixed.


jprberlin:
Can we add city detection as well? I guess this is aggregated enough to not be critical in terms of data privacy//

soldag:
@jprberlin I added the city lookup. This includes loading a CSV provided by Google for mapping city names to so called `Criteria IDs` required by the GA API.

PR #507, opened 2017-12-01
693 nexboard via api implementieren

kremer-io:


kremer-io:
screen shot 2017-12-15 at 10 24 39

janrenz:
ping ...

Langleu:
pong ...

Charle42:
zum testen den Server mit den Branch `258-user-ID-to-Lessons-content-from-user-who-create-it`

kremer-io:
@Charle42 When re-using an existing neXboard, no url is saved. This results to:

screen shot 2017-12-22 at 11 01 26

kremer-io:
@Charle42 When we reuse nexBoards in different topics, we've to specify a new name for the existing board. So the same board is available under different names. @janrenz Do we need some kind of unique checking, for example the neXboard url?

janrenz:
I think no unique checking is needed. lets monitor after rollout for user feedback..

kremer-io:
@Charle42 I'm still getting the same error when re-using boards. Double checked, that all files are up to date.

kremer-io:
@Charle42 When using multiple nexboards in a lesson, there is a problem with the .chosen selector. A random id for each nexboard-block could help. This was already implemented for the CKeditor.

Charle42:
what is the error message? and at which step?

PR #530, opened 2018-01-03
vorstellungsseite schul cloud team

Langleu:
- [x] missing texts
 4 pictures missing, will be added as soon as they are there.
Can still be merged into master.


controllers/team.js
```diff
@@ -0,0 +1,27 @@
+const express = require('express');
+const router = express.Router();
+const authHelper = require('../helpers/authentication');
+
+// Partners
```
pinussilvestrus: vielleicht, ganz vielleicht, könnte man dieses gecopy-paste noch wegmachen 😛

Langleu: war ja noch WIP 😜

janrenz:
screen shot 2018-02-01 at 10 08 01

Langleu:
@janrenz 
a) I will try it out, but problem is that some people have too much text and then the header disappears entirely, but the footer will always be visible.
b) the reason why the nav is overlapping is, that the nav just lies over everything and the content is just moved 240px to the right. Fixed it by adding the padding to the left, same reason why the footer disappears currently in the master branch and your pr as well.

Langleu:
alle RFCs eingebaut .. @janrenz @pinussilvestrus

Langleu:
so dürft wieder

Langleu:
[Vorstellungsseite Schul-Cloud-Team](https://trello.com/c/YVgmn1MG/871-vorstellungsseite-schul-cloud-team)

PR #545, opened 2018-01-18
Allow theming of the frontend

janrenz:
Added Support for custom templates, partials and SCSS


kremer-io:
# Example for SC_THEME=21

screen shot 2018-01-18 at 11 41 00

screen shot 2018-01-18 at 11 42 11

PR #546, opened 2018-01-18
SC-89 ~ nightwatch-frontend tests

Langleu:
just the beginning, trying to combine my test repo with the actual client
- [x] set cookie for new features modal
- [x] clear cookies before new run
- [x] add firefox

These tests were mostly created to test out the permission system.
Another person will later take over the tests and has a big playground for new tests.
It will only run for PR requests as it creates a diff of the PR Branch and Master Branch.

Please use squash & merge

//Update:
Firefox was removed again due to random behavior on travis machine, but it still works perfectly fine on local systems.


Langleu:
Tests behave entirely different on travis (Linux) compared to local testing...

adrianjost:
For me it's working fine under windows. Only problem is that I don't have the material service running on my machine so this test fails.

Langleu:
@adrianjost kind of approved already 21 days ago ^^

janrenz:
why not run it on every master merge?

Langleu:
but that’s pretty much happening?
Just that it looks at the diff between the master and the pr branch and if there are changes in any of the areas that have tests then those are tested, but ofc we can also just run all on a master pr merge

janrenz:
@Langleu I amfoine with the current solution. only add some readme

Langleu:
@janrenz added some README, please check.

PR #548, opened 2018-01-18
theme improvements

adrianjost:
- gulp is watching theme folder
- overwrite src files with theme files if available (could be improved in gulp)


janrenz:
Great work!

PR #554, opened 2018-01-31
changed max-width in homework area

adrianjost:


adrianjost:
[Meppen - use full width for homeworks](https://trello.com/c/PYoQT2sF/901-meppen-use-full-width-for-homeworks)

PR #556, opened 2018-02-01
Jr/footer

janrenz:
just made the footer smaller
screen shot 2018-02-01 at 09 14 57


janrenz:
We should also make it even left align or centered in the content area..

PR #573, opened 2018-02-19
Feature/intro

Trunken:
IntroJs Guide für die Didacta


Dockerfile
```diff
@@ -15,4 +15,4 @@ USER node
 WORKDIR /home/node/app
 
 EXPOSE 3100
-CMD ["node", "bin/www"]
```
janrenz: Does this has any performance impact in production

PR #574, opened 2018-02-19
911 fix sc card menu

adrianjost:
image

to

image


adrianjost:
[Fix sc-card-menu](https://trello.com/c/QFFi0ffA/911-fix-sc-card-menu)

PR #580, opened 2018-02-20
color Picker fallback

adrianjost:
image


adrianjost:
[Add polyfill for colorpicker](https://trello.com/c/OhOD9AIL/915-add-polyfill-for-colorpicker)

PR #588, opened 2018-02-25
fix sass error

adrianjost:
If you changed something in static/styles/homework/homework.scss there was an error like:
`Error: Invalid CSS after "...-repeat center;": expected 1 selector or at-rule, was "﻿@import '../lib/co"`
It seems like this is an node issue because the error looks almost identical to this ticket.
~ [Ticket: https://github.com/fourseven/meteor-scss/issues/150](https://github.com/fourseven/meteor-scss/issues/150)

Anyway, if you add an comment at the end of the `theme/` style files the error doesn't appear. So gulp is now adding this comment to any injected style file.


gulpfile.js
```diff
@@ -81,7 +81,7 @@ gulp.task('styles', () => {
     var themeFile = `./theme/${themeName()}/style.scss`;
     beginPipe('./static/styles/**/*.{css,sass,scss}')
         .pipe(sassGrapher.ancestors())
-        .pipe(header(fs.readFileSync(themeFile, 'utf8')))
+        .pipe(header(fs.readFileSync(themeFile, 'utf8')+"// don't remove me! \n")) // READ: https://github.com/schul-cloud/schulcloud-client/pull/588
```
pinussilvestrus: image

The `homework.scss` still make problems on my machine..

adrianjost:
okay I tried a lot of things. Neither works... :( 
Can anyone else try to solve this issue?

My current guess and result is the following:
1. The problem only occurs if sassGrapher includes any new files to the stream
2. If you move all the variables from the themeFile to the colors.scss file everything works fine.

janrenz:
Maybe something like line endings or invisible characters ?

adrianjost:
yup, maybe. But I now tried 
1. to use add universal line endings `\r\n`
2. converted all files again to utf-8 (even when they already where utf-8)
3. changed all import paths to `./` for files in the same directory
4. When I delete the complete homework style folder and removing any reference to it I can't produce the error. So the failure should be in any of these few files...

=======================
5. I regenerated new files and copied the old content into them. At first look, this seems to solve the issue. Please test :)

adrianjost:
Seems like there was an encoding error when "UTF-8 without BOM"-files should be combined with the theme file that was read as UTF-8. I added an Test to ESLint that should fix this issue in future automaticly.

adrianjost:
TODO: don't use SassGrapher on first run! It will take realy long and isn't needed.

pinussilvestrus:
image
Ist das so gewollt?

adrianjost:
nope... hab ich wohl was in die falsche Richtung gemerged :/

adrianjost:
@pinussilvestrus sollte jetzt wieder richtig sein

PR #598, opened 2018-03-15
[WIP] SC-86 ~ Aufgabe erstellen bug wenn kein datum gesetzt

adrianjost:


controllers/homework.js
```diff
@@ -79,6 +79,12 @@ const getCreateHandler = (service) => {
             if ((!req.body.lessonId) || (req.body.lessonId && req.body.lessonId.length <= 2)) {
                 req.body.lessonId = null;
             }
+            if(req.body.dueDate == "__.__.____ __:__"){
```
janrenz: can we make this more generci, for example if parsing the date fails this will be undefined

adrianjost: You are right, I've checked again and saw, that we already have such an fallback. But for any reason the fallback is not working on the production system. If I create an task on my local test server with the master branch without any date set everything works fine. If I do the same thing on schul-cloud.org I get the alert "Das Beginndatum muss vor dem Abgabedatum liegen!".
Do you have any idea why?

janrenz:
OK, so this is just cleaning up a new line and removing the console.log ?

adrianjost:
jup, I haven't found the real bug. For now, I marked this pull as WIP.

Do you have any idea why the [existing fallback](https://github.com/schul-cloud/schulcloud-client/blob/46e88eb46650474ebed94ec93bcc51a572ddaa45/controllers/homework.js#L92) isn't working on the production system? On my local development setup everything works fine.

janrenz:
Different locale settings?

adrianjost:
What do you mean?

---------------

Currently I'm guessing that the problem is in [line 85](https://github.com/schul-cloud/schulcloud-client/blob/46e88eb46650474ebed94ec93bcc51a572ddaa45/controllers/homework.js#L85) and [line 89](https://github.com/schul-cloud/schulcloud-client/blob/46e88eb46650474ebed94ec93bcc51a572ddaa45/controllers/homework.js#L89). I think that moment isn't acting the same on my device as on the server environment when parsing an invalid date. I added a specific fallback value and hope this works. Maybe someone else can try to reproduce the original error on his device and then test if this fix works? I can only guess what's wrong because I can't reproduce the error on my machine.

Maybe @pinussilvestrus ? Here is the [ticketlink](https://ticketsystem.schul-cloud.org/browse/SC-86)

janrenz:
different node or moment version?

adrianjost:
what node version are we using in production?

Lars found out that the problem only occurs on linux.

pinussilvestrus:
@adrianjost any updates regarding to this?

adrianjost:
@pinussilvestrus I'm a Windows user and can't reproduce this bug. So it should be checked by any Linux user. For now I have no idea why this happens.

adrianjost:
I tested it right now and the bug seems to be solved.

PR #608, opened 2018-03-28
[WIP] SC-54 ~ disable tooltips on mobile devices

adrianjost:
May this works. Need to be tested!


adrianjost:
better fix in another branch

PR #611, opened 2018-03-29
SC-75 ~ add Schul-Cloud Component

adrianjost:
ToDo:
1. Link not editable
2. check that link is schul-cloud intern
3. check for ENV variables for button title and url check

[Ticket](https://ticketsystem.schul-cloud.org/browse/SC-75)


pinussilvestrus:
@adrianjost 
1. implement the updateHandler
2. include an input-pattern for validation. When hitting the save button, the form is displaying an standard invalid-error when the link is not beginning with the current system`s base-url
3. made it system-independent by just naming it "Interner Link"

Langleu:
@adrianjost @pinussilvestrus is this ticket still in WIP or not?
Couldn't really find the ENV variables mentioned in the TODO from @adrianjost , but @pinussilvestrus changed it nevertheless to not WIP anymore.

pinussilvestrus:
We do not need any Environment Variables here, just because there shouldn't be any schulcloud-related anymore.

Langleu:
I'd prefer to have a nice and clear in the first try and not the 23rd, but that is just my cup of tea 🍵

adrianjost:
agree with @Langleu . Teachers often do crazy stuff so this new tool should be pretty safe to use before merging.

janrenz:
The external URL checke is up and running and could be used here. Maybe a generic hook would make sense here.

pinussilvestrus:
Can you provide more information about the external URL checker? Do you mean the https://github.com/schul-cloud/antivirus_check_service?

pinussilvestrus:
@Langleu @janrenz as far as I can summarize, we should implement to checks:

- [ ] blacklist-checker for several critical (internal) urls in the `topic-controller `
  - [ ] The Topic itself, just kinda made the whole system super slow at me.
  - [ ] Personal links like my files, my account settings ... (admin area)
  - [ ]  `/edit`, `/new` - pages

I would prefer doing this in the frontend-controller, just because it's the links are also system related. Or why would you do this in a hook-function?

I don't know whether an antivirus-check-hook would make sense since it's only possible to insert internal-schul-cloud/internal-links. However if you can show me one example, where critical virus-content can be included in an internal page, we can also discuss this :)

pinussilvestrus:
@adrianjost @Langleu the pattern is implemented, but its very very slow (the input-field often does not react for half a second): https://github.com/schul-cloud/schulcloud-client/blob/063867b1cedb8a59fa34377ddf4ba32e657b3876/static/scripts/topicEdit.js#L757

Any ideas to make it more efficient?

adrianjost:
@pinussilvestrus I will take a look on it the next days. Maybe I can find an fix

adrianjost:
I was able to reproduce your problem with an slow query. My page even froze completly. I found out, that the problem seems to be that you use a lot of wildcard matches. If you combine every case into one simple matching or group it works perfectly for me.
```
return `(${window.location.origin})(?!.*\/(edit|new|add|files\/my|account|administration|topics\/${topicId})).*`
```
I know it isn't as easy to read as before but it should be way faster. Feel free to test.

BTW. I would also add this pattern as an check in the server like @Langleu mentioned. Some browsers do not support these pattern check and you know, some teachers are using incredible old browser. We may have an performance problem when someone is able to enter the topic itself here because everyone viewing this ends in an loop. So better add an server hook for that.

pinussilvestrus:
@adrianjost Big thanks to you!! :+1: I will Test it later on. And yes I will do another check on Server-Side (but rather in the Controller), it's the last checkbox in my todo-list few comments ago :)

pinussilvestrus:
@adrianjost I also implemented a regex-check in the controller now. Discussed it with @Langleu that the controller is a better place for it, since we can't check the client's url in the server. 
In future there will be a [client check](uhttps://github.com/schul-cloud/schulcloud-server/pull/297) in the server, so it will be not possible to inject these critical url's via the raw api. 

Have fun in reviewing again! 🙌

Langleu:
mir ist da noch ne Kleinigkeit aufgefallen, aber nix schlimmes.
Sowas wie `http://localhost:3100/files/file?download=1&file=users/0000d231816abba584714c9e/groovy.jpg` sollte nicht möglich sein, also nur euren Regex erweitern. bzw. `http://localhost:3100/files/file?file=users/0000d231816abba584714c9e/groovy.jpg`

falls irgendwer auf die Idee kommt, eine Datei direkt zu verlinken

Langleu:
vielleicht sollte man auch die topics komplett verbieten?
Ich weiß nicht ob ihr von früher aus der Schule noch diese maze ordner kennt? Also Ordner A zeigt auf B, B zeigt auf C und C wieder auf A und irgendwo gab es ne Kombi wo man zu dem Ordner hinkam wo an hin wollte?
Der Spaß lässt sich auch mit den Topics rekonstruieren ^^ Inception like.

image

Also nur mal als Einwand, weil ja auch Schüler Themen erstellen können ^^, aber ist ja auch nur ne kleine Änderung.

PR #612, opened 2018-04-03
Update extended_footer.hbs

hannoterhaseborg:
mailadress fixed


theme/n21/views/lib/extended_footer.hbs
```diff
@@ -1,10 +1,10 @@
-<footer>
+oter>
```
adrianjost: <footer>

PR #625, opened 2018-04-06
re-add login form

kremer-io:
Login Form was missing.

image


kremer-io:
Now it's fixed

image

PR #639, opened 2018-04-18
Remove custom tooltip that annoyes on ios and touch devices

janrenz:


adrianjost:
What about: https://github.com/schul-cloud/schulcloud-client/pull/608 ?

janrenz:
@adrianjost detection seems not reliable based on what I read. so why not get rid. the title attribute will be displayed on normal browsers.

PR #661, opened 2018-04-26
fix footer in safari, fixes course add button

janrenz:


static/styles/lib/base.scss
```diff
@@ -89,6 +89,7 @@ footer.small-footer {
 	clear: both;
 	padding: 30px 0px;
 	font-size: 0.9em;
+	margin-bottom: 5px;
```
adrianjost: also substract 5px to the min-height calculation here https://github.com/schul-cloud/schulcloud-client/blob/master/static/styles/lib/loggedin.scss#L238

PR #672, opened 2018-05-14
Zwei Wörter zu viel ....

hannoterhaseborg:


schurigh:
@kremer-io can we merge this with the next n21 update?

PR #690, opened 2018-05-25
Optimize production docker

janrenz:
- Replace duplicates files by symlinks
- Only rebuild files with other themes that are affected by theming

need to consider of maybe having two docker versions, one for production, one for dev


janrenz:
Setting on docker hub is done.

janrenz:
@kremer-io ping.. basically this just the same docker files as before , just old dev and new prod..

kremer-io:
@janrenz I've updated the docker-compose to the the new Dockerfile and removed the development version

PR #692, opened 2018-05-28
add readme how to set timezone

adrianjost:
nothing to do here. Simply set the enviroment variable where you run the client.

https://stackoverflow.com/questions/8083410/how-can-i-set-the-default-timezone-in-node-js


janrenz:
Still needs to be fixed in Prod? @kremer-io

kremer-io:
@janrenz I'll test this on prod.

kremer-io:
@adrianjost Many many thx. Is working!

PR #698, opened 2018-05-31
second poll and preferences

Langleu:
finally make use of preferences.
Introduce the frontend route `/account/preferences`
Post Data as followed : `{ attribute: { key: "key", value: value } }` and it will be written into the preferences.
Preferences are automatically loaded in a hidden object in the loggedin.hbs meaning the preferences are everywhere available for scripts via jquery.

### DIFF
--- a/controllers/account.js
+++ b/controllers/account.js
@@ -77,4 +77,22 @@ router.get('/user', function (req, res, next) {
     res.json(res.locals.currentUser);
 });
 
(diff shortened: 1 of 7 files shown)


janrenz:
@Langleu can we then move the releases info persistence to this as well?

Langleu:
sure @janrenz :+1: , I'll do it soon.

PR #699, opened 2018-05-31
SC-240 - add overflow to home menu

adrianjost:
image
image


schurigh:
Hmm... not sure this is the best way. Its not obvious enough that there are more menu items available if you see this:
image
Whats with the good old hamburger menu button mit dropdown on click?

adrianjost:
@schurigh
Hamburger Icon seems to be overkill and in general it is categorized as bad UX for the main navigation. Read here: https://medium.theuxblog.com/should-we-really-kill-the-hamburger-button-626090cd0cd

I was just following the Material Guidelines with this commit. https://material.io/design/components/tabs.html#scrollable-tabs

adrianjost:
image
updated to follow these guideline

schurigh:
Thanks for all the details. I'm not that big of a mobile browser user, instead I focus on apps. So I wasn't aware of that UI trend/change and yeah, it is official Material Design, but still, that looks kinda wrong as I never saw/used a scrollable horizontal nav before:
image
I will let the team have a look at the change on monday.

adrianjost:
Here is a version with a small hint for the user.

overflow

BTW: Over the weekend I noticed that it's more widely used than I thought. For example: 
Github
image
Google
image

schurigh:
I have showed it to four coworkers here and there are three main comments:
- hamburger menu is not dead (as you can see on your last screenshots) and proved as a good solution for most of us
- visualization of scrollable nav is needed (scrollbar, gradient fade or icon) to support it
- both combined might work (as google and github did as well) (my fav)
Someone have to decide ;)

janrenz:
I would keep it in sync with the internal nav...

schurigh:
actual status? final || continue || discard?

adrianjost:
discard

PR #703, opened 2018-06-05
Save ReleaseDate as Pref instead of Cookie

Langleu:
needs https://github.com/schul-cloud/schulcloud-server/pull/309


janrenz:
Locks great. But maybe we can even go one step further: We could read both preferences and release date on the server and then only trigger the client to show release info once. This would avoid the call to /help/releases on every page..

Langleu:
if we had stuck to React and websockets this would have been an easy one.
I'll look what I can do

janrenz:
also it would be ok to do this only as part of the login action

Langleu:
sure thing. I'll try some stuff out.

Langleu:
@janrenz changed some stuff and it now would use websockets.
The User logs in and the server checks whether the user has seen the newest release, if not it triggers an event via websocket.
Will probably still move it to the dashboard, as it's overkill to check it every time.

janrenz:
plus one for moving it to dashboard

Langleu:
in a frontend script?

janrenz:
should be stored in a env and then somehow passed to the script.. otherwise this wouldn't run in production or am I overseeing something?

Langleu:
changed it, env variable will be to much of a mess to get it into the frontend.

static/scripts/socket.js
```diff
@@ -1,5 +1,6 @@
 $(document).ready(function () {
-    let socket = io('http://localhost:3030');
+    let url = window.location.href.includes('localhost') ? 'http://localhost:3030' : 'https://schul-cloud.org:8080';
```
schurigh: I'm running http://127.0.0.1:3100/ locally ;)

Langleu: switched the cases, shouldn't matter anymore.

Langleu:
allet klar, mach ich dann beim nächsten Mal.

Langleu:
@janrenz done.

helpers/handlebars/middleware.js
```diff
@@ -33,6 +33,7 @@ const makeActive = (items, currentUrl) => {
 
 module.exports = (req, res, next) => {
     res.locals.themeTitle = process.env.SC_NAV_TITLE || 'Schul-Cloud';
+    res.locals.backendUrl = process.env.BACKEND_URL || 'http://localhost:3030';
```
schurigh: Socket.io won't work if you test on http://127.0.0.1 instead of localhost but that might be unusual

Langleu:
changed it to the public backend url as the normal backend url is used for docker referencing meaning server:3030 aint gonna work for any connections

janrenz:
on mobile we dont show the release info pop up

PR #705, opened 2018-06-07
start adding functionality to searchbar or hide it

Langleu:


schurigh:
Perhaps I am overlooking things, but: /courses already lists all courses of that user, 100%/maximum, the search can just limit/filter/reduce the existing amount. If so, I would like to kill the need to submit by pressing enter and additional database stuff and just do it live-JS:
$(search).on(keypress($(coursetitle:not(:contains($(search).val())) or so

schurigh:
Another thing: Would it be better to add a few showSearch:true where search is needed instead of many many hideSearch that one has to think of in any new route? Standard: no search bar and add it just if dev knowingly add showSearch.

static/scripts/loggedin.js
```diff
@@ -250,4 +250,21 @@ $(document).ready(function () {
...
+    jQuery.expr[":"].Contains = jQuery.expr.createPseudo(function(arg) {
+        return function( elem ) {
+            return jQuery(elem).text().toUpperCase().indexOf(arg.toUpperCase()) >= 0;
+        };
+    });
+    // js course search/filter
+    $("input.js-search").on("keyup", e => {
+        if(e.key === "Escape") $(e.target).val("");
+        if(e.key === "Unidentified") {
+            e.preventDefault();
```
adrianjost: use `return false` to do both at once ([article](https://stackoverflow.com/questions/1357118/event-preventdefault-vs-return-false))

`return (e.key === "Unidentified")` at the end

adrianjost:
I doesn't have time to test it today but will do so during the next days

schurigh:
@janrenz True, didnt know that pagination is used here. Hmm... perhaps back to db query then.

Langleu:
I actually like the live search and would keep it as it is.
@schurigh didn't disable the db search meaning you can now combine the both of them.
You can search for something and afterwards still specify more precisely what you mean.

schurigh:
Multiple things I want to add:
1. There is no paging visible if there are more than 25 courses. Only 25 are shown but there are no "pages" to go to next courses.
2. Live search may confuse if user has more than 25 courses, only 25 are shown, he inputs smth and courses gets filteres, but only the 25 first shown courses. Is it obvious enough that [enter] in that search box starts a new search and might reveal other courses that weren't shown before?
3. If the user has pressed [enter] and only a subset of courses is shown, how to "unfilter"? Search for empty string (just press [enter]) or reload page is possible but not so obvious for every user I think. Maybe add like a little lable besides search field with a delete/X button to remove that search/reset search?
Or we just remove paging/25-filter on courses db find.

Langleu:
1. Will extend the $limit to 50 for now and whoever will do the course archive and all that will handle it then.
2/3. As courses will be extended to 50 courses for now, I would probably leave the live search as it is and remove the other option.
For news on the other hand I'll remove the live search,  as pagination is already included and live search has no advantage over db query.

Langleu:
@schurigh changed it to what I mentioned.

schurigh:
Little remarks: 
1. Pressing [enter] still fires up a search with "?q=[searchterm]" url appendix but no search is done. This was not intended, right? 
2. Just to be clear: there's still no visible paging, right? You upped to 75 results, thats a lot. But IF (rare, i know) someone have more than 75 courses (Schuljahreswechsel might be a moment because of duplicating courses), theres no way to see your 76+, especially if [enter]-search-to-filter is not working.

Langleu:
1. done
2. as I already said that whoever will do the archive and all should deal with pagination, as he/she has to redo the whole design anyway. Also it doesn't directly concern this ticket. You should probably add your concerns to the ticket for the archive/schuljahreswechsel, if it already exists. Probably when that is implemented we could go back to the db search, but for now the livesearch will do.

PR #707, opened 2018-06-12
Nexboard adjustments

Langleu:
use adjusted lib


janrenz:
Proxy for a next iteration sounds fine... let’s keep this in mind

Langleu:
ty @pinussilvestrus 
Would also support the proxy service :+1:

PR #714, opened 2018-06-19
remove duplicate eventListeners

adrianjost:


schurigh:
Next time, please provide even a little bit of information which problem/behaviour this code should fix and so on, that would support the tests ;)

PR #736, opened 2018-07-19
Consent-App, Depseudo, Activity Log

dominikglandorf:


janrenz:
This is not needed anymore

PR #741, opened 2018-07-26
Feature/improve watch mode building time

georgberecz:
**Problem**
This PR tackles two problems:
1) Event Listeners from `loggedin.js` were attached multiple times on elements.
2) Changing and saving dependencies would not retrigger a build of dependent files (e.g. `a.js` imports a function from `b.js`: If something was changed in `b.js`, `a.js` needs to be rebuilt as well). That was not a problem before, because without webpack there was no such thing as a dependency tree and `gulp watch` would just rebuild everything that was changed.

**Solution**
1) I moved the function into a helper module. It actually does not make sense that `loggedin.js` exports `softNavigate`. By doing so, `loggedin.js` is only imported a single time and event listeners are attached as they should.
2) I took a deep dive into webpack internals and wrote a webpack-plugin that solves this problem by caching and saving dependencies for entrypoints. Currently, simply the modification date is used to determine whether a file changed and a rebuild has to occur. More information can be found here: https://github.com/georgberecz/rebuild-changed-entrypoints-webpack-plugin


georgberecz:
Is there any reason why TravisCI relies on Node 7? (This is the reason the builds are failing). Can we upgrade to the current LTS 8.11.3?

Langleu:
1. worked
2. changing any script triggers gulp task `scripts` which would rebuild all scripts?
---
did you ever try out theming? `SC_THEME=n21` it kinda broke the whole building process for me and now not even `gulp watch` finishes correctly causing missing files.

Langleu:
Thanks for explaining!

Yes exactly, first built it, then cleared it and ran it with `SC_THEME=n21` again.
Now both are working properly again.

PR #752, opened 2018-08-03
SC-356-deny-access-to-welcome-files-for-demo-accounts

schurigh:
Ticket: https://ticketsystem.schul-cloud.org/browse/SC-356
Deny access to Hilfebereich -> Dokumente aus dem Willkommensordner for Demo Accounts with permission check


controllers/help.js
```diff
@@ -44,17 +45,27 @@ faq.ssoFAQ.map(faq => {
...
-    let documents = faq.documents;
-
-    documents[0].content = converter.makeHtml(documents[0].content);
-
-    res.render('help/sso-faq', {
-        faq: documents,
-        title: "Dokumente des Willkommensordners zum Download"
-    });
+    // check a random permission that demo users dont have to detect demo accounts
+    let access = !!permissionHelper.userHasPermission(res.locals.currentUser, 'FEDERALSTATE_VIEW');
```
pinussilvestrus: I think this `!!` operator is only necessary if there could be `undefined` or `null` as return-value. Since the `userHasPermission`-function only returns true or false, it isn't necessary in my opinion.

PR #817, opened 2018-10-08
SC-485-Sonderzeichen bug

KatharinaBlass:
ticket: https://ticketsystem.schul-cloud.org/browse/SC-485


KatharinaBlass:
I have changed your 2 things and some more, which I seemingly forgot :grimacing:

PR #835, opened 2018-10-16
SC-463-icons für externe Links

KatharinaBlass:
ticket: https://ticketsystem.schul-cloud.org/browse/SC-463
nur für den Kursbereich

### DIFF
--- a/static/styles/lib/base.scss
+++ b/static/styles/lib/base.scss
@@ -308,7 +308,7 @@ form.inline {
 //the following adds fa-share-square-o for each external link except images
 .section-course .content-block{
 	a:not( [href*='https://schul-cloud.org'] ):not( [href^='#'] ):not( [href^='/'] ){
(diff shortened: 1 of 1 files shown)


static/styles/lib/base.scss
```diff
@@ -304,3 +304,16 @@ form.inline {
 .position-absolute{
 	position: absolute !important;
 }
+
+//the following adds fa-share-square-o for each external link except images
+.section-course .content-block{
+	a:not( [href*='https://schul-cloud.org'] ):not( [href^='#'] ):not( [href^='/'] ){
+		&:not( [href$='.jpg'] ):not( [href$='.png'] ):not( [href$='.jpeg'] ):not( [href$='.gif'] ){
```
schurigh: Can we embed .tif and .bmp into courses? If so, add this here. If not, it's fine.

KatharinaBlass: I added .tif and .bmp, but i don't know if we will need it

schurigh: Thats why I asked if they even are embedable in courses. If not, we dont need it. I will check it...

PR #836, opened 2018-10-17
added basic http2 handling

adrianjost:
TODO:
- store cert securely and less static (files for dev & prod)
- redirect (localhost) from http to https?
- add readme for dev-certificate usage


janrenz:
ok, we need to discuss how this will work in our infrastructure, https stuff will hapen in nginx before the server. also we could make this optional for the dev env ..
Also @tofixx should be involved here

janrenz:
@kremer-io should we test it in NginX?

tofixx:
we should but maybe let's run my tests for one week with http1 that we may compare the loading times

tofixx:
@kremer-io think we could switch nginx to http2

PR #795, opened 2018-09-13
SC-161-upgrade-webcomponents

malcm:
necessary webcomponents of m.e.i.n.e.l and video player are not compatible. Change in video-player and or m.e.i.n.e.l would be necessary (won't fix)

@janrenz  Possible update of video-player to version 2.2.2 (also would use webcomponents-bundle.js and fix this issue). But gulp building process destroy file video-player.js. Any suggestion?

PR #845, opened 2018-10-22
Tobias/local vendors added

tofixx:
moved external ressources to local vendors directory


tofixx:
however, gulp seems to not copy all files using watch

janrenz:
vendor stuff should not be monitored by gulp watch. I think there is already some code or a PR adressing this. Also do we take crae that those librariers are only loaded where needed?

tofixx:
ok. however the library is always imported when logged in

janrenz:
i think it only need to be loaded on pages where the rich editor is used. The next iteration of the editor should support lazy loading ...

adrianjost:
@tofixx It's working now. But do you have any ideas what we can do to speed up the process? gulp is running for a minute in my machine to build mathjax. For local development it's okay because we are caching, but the docker build time for our production server will increase a lot I think. And as far as I understand docker, we are running gulp on every restart?
@janrenz could this be a problem?

janrenz:
No, gulp is just beeing used on build time, i refactored this a while ago

Michael-Janke:
> @tofixx It's working now. But do you have any ideas what we can do to speed up the process? gulp is running for a minute in my machine to build mathjax. For local development it's okay because we are caching, but the docker build time for our production server will increase a lot I think. And as far as I understand docker, we are running gulp on every restart?
> @janrenz could this be a problem?

using webpack 😄 
it supports incremental builds, multi threading and bundling by design.
and it helps to only include things you really need. (And use npm instead of hard copies)

To be fair, also gulp4 is faster than gulp3

tofixx:
@Michael-Janke the target of this pr was to move external vendors to a local path. Feel free to open a new pr that focuses on optimization of gulp.

malcm:
@janrenz @tofixx I just reconized that the PR #795 was closed since this PR should include all changes. But it misses updating the video-player from version 1.3.2 to 2.2.2 and also misses updating m.e.i.n.e.l
Do you add this @tofixx or should I do it?

malcm:
ok. I just didn't got why #795 was closed if it was not completly added here. I will take care of it in the upcoming days

adrianjost:
Sorry @malcm I just overlooked these changes.

PR #853, opened 2018-10-30
no more require BroadCastChannel API for calendar event updates

tofixx:
changed messages to use postMessage interface instead

### DIFF
--- a/static/scripts/calendar.js
+++ b/static/scripts/calendar.js
@@ -55,20 +55,22 @@ $(document).ready(function () {
         editable: false,
         timezone: 'UTC',
         events: function (start, end, timezone, callback) {
+            if('serviceWorker' in navigator){
+                navigator.serviceWorker.addEventListener('message', function (event) {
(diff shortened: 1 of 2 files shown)


static/scripts/calendar.js
```diff
@@ -55,20 +55,22 @@ $(document).ready(function () {
         editable: false,
         timezone: 'UTC',
         events: function (start, end, timezone, callback) {
+            if('serviceWorker' in navigator){
+                navigator.serviceWorker.addEventListener('message', function (event) {
+                    if (event.origin !== 'https://schul-cloud.org' || 'http://localhost:3100')
```
janrenz: this should come from the cofig, as we run on other domains as well

tofixx: maybe we just can remove that check. this is a litte security check but the check for data.tag should be enough as well

janrenz: i am ok with removing

tofixx: solved: we can auto detect this

kremer-io:
@janrenz @tofixx pls always use squash and merge

PR #863, opened 2018-11-06
cg/UpdateTeamPage

CarolinGoerke:


schurigh:
@CarolinGoerke You should add a picture of yourself like the other team members so that it doesnt look like 
image

PR #872, opened 2018-11-07
fix topic edit button link

schurigh:
don't know why this was changed from /courses to /{{context}} but reverting fixes the problem. maybe check if {{context}} was needed for something else?


schurigh:
Review by @smyth64

PR #875, opened 2018-11-07
align header style of topics and tasks in course detail page

janrenz:
screenshot 2018-11-07 at 03 41 43

### DIFF
--- a/static/styles/courses/course.scss
+++ b/static/styles/courses/course.scss
@@ -126,7 +126,11 @@
     display: none;
   }
 }
-
+.course-task-header {
+  padding: 0.5rem 0.75rem;
+  background-color: $primaryColor;
+  color: $colorWhite;
+}
 .card-topic {
   cursor: pointer;
 
--- a/views/courses/components/tasks.hbs
+++ b/views/courses/components/tasks.hbs
@@ -1,10 +1,10 @@
 <div class="row">
   <div class="col-sm-12 container">
-    <h4>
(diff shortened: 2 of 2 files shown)


static/styles/lib/colors.scss
```diff
@@ -1,2 +1,3 @@
 $colorBeige: #e6e2e2;
 $colorGrey: rgba(232, 232, 232, 0.5);
+$colorDarkerGrey: #555555;
```
dominikbruechner: Die Farben fürs StandardTheme sind jetzt alle in theme/default/style.scss - sollte die auch dahin?

janrenz: ist ja ne Farbe und keine sematische Zuordnung .. also eher nein würde ich sagen

janrenz:
as dicussed: header will be hidden if there is nothing to filter ...

janrenz:
@dominikbruechner i do not like using a generall permission, why not use something from within the view for deciding like {{#if ../myhomeworks.length }}

dominikbruechner:
> @dominikbruechner i do not like using a generall permission. This might easily break once we change the permissions. Why not use something from within the view for deciding like {{#if ../myhomeworks.length }}

I am going to rethink the permission part. Students right now generally have permission to create private homework in this view, which does not fit into the context of this particular view. 
Hiding the lower part if it is with myhomeworks < 1 makes sense, although I'd generally prefer to not enable students to create tasks in this view...

PR #877, opened 2018-11-08
remove additional parameters from url

tofixx:
optimized cleanup for analytics logging


konstantin-sc:
Requeste mal einen einen anderen Reviewer, ich hab doch keine Ahnung und winke nur durch ;_)

janrenz:
will this work with nested urls? maybe add some test for the getPath Function...

tofixx:
> will this work with nested urls? maybe add some test for the getPath Function...

this will cut an url on first occurence of `#` or `?`. how do you mean nested?

janrenz:
Ok, we then should monitor this when introducing new routes

PR #879, opened 2018-11-08
Fix/gulp css

tofixx:
- seems, that #854 resulted in failures compiling the scss files. please check, `gulp clear && gulp` will result now in compiling all files correctly
- missing return statements reuired to detect when tasks finished in gulp were added too.
- first run detection has been updated after styles task

image

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -97,8 +97,9 @@ gulp.task('styles', () => {
     .pipe(filelog("PROCESS: "))
     .pipe(sourcemaps.init())
     .pipe(sass({
-      sourceMap: true
-    }))
+      sourceMap: true,
+      includePaths: ['node_modules']
+    }).on('error', sass.logError))
     .pipe(postcss([
       autoprefixer({
         browsers: ['last 3 version']
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -97,8 +97,7 @@ gulp.task('styles', () => {
     .pipe(filelog("PROCESS: "))
     .pipe(sourcemaps.init())
     .pipe(sass({
-      sourceMap: true,
-      includePaths: ['node_modules']
+      sourceMap: true
     }))
     .pipe(postcss([
       autoprefixer({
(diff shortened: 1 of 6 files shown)


schurigh:
I didn't had any CSS problems the last days. May someone review who had.

static/styles/lib/base.scss
```diff
@@ -4,7 +4,7 @@
 @import "./tether/tether.min.css";
 @import "./bootstrap/scss/bootstrap-flex";
 $fa-font-path: "/vendor-optimized/font-awesome/fonts";
-@import "font-awesome/scss/font-awesome.scss";
+@import "../../../node_modules/font-awesome/scss/font-awesome.scss";
```
Michael-Janke: ../../ is not a proper way to get to node_modules
why changing this?

Michael-Janke: try
@import "~font-awesome/scss/font-awesome";
(i have no time sry)

tofixx: it should if webpack would be used for that, tried some variants but they didn't work.

Michael-Janke: You are right!
https://stackoverflow.com/questions/41647984/importing-sass-partials-from-node-modules
with importPaths it should work

@adrianjost maybe you can doublecheck

gulpfile.js
```diff
@@ -91,14 +91,13 @@ sassGrapher.init('./static/styles/', {
...
   var themeFile = `./theme/${themeName()}/style.scss`;
-  beginPipe('./static/styles/**/*.{css,sass,scss}')
+  return beginPipe('./static/styles/**/*.{css,sass,scss}')
     .pipe(gulpif(!firstRun, sassGrapher.ancestors()))
     .pipe(header(fs.readFileSync(themeFile, 'utf8')))
     .pipe(filelog("PROCESS: "))
     .pipe(sourcemaps.init())
     .pipe(sass({
-      sourceMap: true,
-      includePaths: ['node_modules']
```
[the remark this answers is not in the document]
tofixx: the initial issue was, that the css files were not build without the output of any error message. not sure how it's working now :)

PR #885, opened 2018-11-09
Show counter in course tab so user know if its worth clicking

janrenz:
screenshot 2018-11-09 at 10 38 43


adrianjost:
@dominikbruechner I'm unsure if we should show the badge if there aren't any items. So if itemcount == 0 I would recommend to hide the badge to reduce the visual stress for the user with one alert badge here, one here and so on.

janrenz:
I would be ok with hiding the badge, so does iOS

PR #891, opened 2018-11-13
only take logs if logged in

tofixx:
do not take logs if logged out, what fails currently because testUserGroup element only in DOM if logged in


janrenz:
we might also consider disabl elogging for demo accounts as well

PR #900, opened 2018-11-16
SC-538 make card transparent when not visible to students

dominikbruechner:
- make card transparent when not visible to students

- minor optical improvement to link:hover effects
screenshot 2018-11-16 at 10 47 48


adrianjost:
I think the contrast still isnt hard enough. Have you tried the grayscale filter like we use it for "archiving an homework"

PR #921, opened 2018-11-29
added johannes to team page

jaunruh:
added johannes to team page


schurigh:
@jaunruh You need to add a picture as well...
image

PR #935, opened 2018-12-05
made context switch course - files more usable

dominikbruechner:
+ added footer margin in mobile
+ made context switch course - files more usable
screenshot 2018-12-05 at 16 10 01

screenshot 2018-12-05 at 16 10 27

screenshot 2018-12-05 at 16 13 54


static/styles/lib/base.scss
```diff
@@ -160,6 +160,7 @@ footer {
 
 		@include media-breakpoint-down(xs) {
 			background: $colorBeige;
+			margin-top: 50px;
```
janrenz: this coudl be moved to a constant

dominikbruechner: like $margin-top-50 ?

janrenz: more than related to the grid system margin-top: $2gridUnits

adrianjost: Why? It's that specific here. 
In addition, there is already bootstrap stuff like `mt-1`, `mt-2`, ...

schurigh: I would not do that, constant for a specific margin-top value? nah...

janrenz: If we follow a global grid there are no specific margins. But its just a proposal, can easily changed later. and maybe something that we need to discuss with mona

PR #937, opened 2018-12-05
added link to course files

dominikbruechner:
Analytics showed that there should be a link between assignments and course files. here it is
screenshot 2018-12-05 at 16 56 46


views/homework/extended.hbs
```diff
@@ -73,10 +73,20 @@
             {{/userHasPermission}}
         {{/inArray}}
         {{#if courseId._id}}
-            <a class="btn btn-secondary" href="/courses/{{courseId._id}}" alt="zum Kurs der Aufgabe">zum Kurs</a>
+            <a href="/files/courses/{{courseId._id}}" class="btn btn-add btn-secondary">
```
janrenz: do we need a permission check here?

dominikbruechner: there is a check if a course exists
otherwise, everybody can access the course files, right?

PR #952, opened 2018-12-12
UX-7 add new color picker to courses

dominikbruechner:
- to restrict selectable colors for future color palette features
- default #ACACAC (other server PR reflects to server model)
- it's no beauty but works for now till the bigger rework with onboarding happens...
screenshot 2018-12-12 at 16 56 10

### DIFF
--- a/controllers/courses.js
+++ b/controllers/courses.js
@@ -131,6 +131,9 @@ const editCourseHandler = (req, res, next) => {
             course.teacherIds.push(res.locals.currentUser);
         }
 
+        // populate colors
+          const colors = ["#ACACAC", "#D4AF37", "#00E5FF", "#1DE9B6", "#546E7A", "#FFC400", "#BCAAA4", "#FF4081", "#FFEE58"];
+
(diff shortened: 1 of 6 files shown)


janrenz:
And the yellow must take  care that the the text is Black for accessibility

dominikbruechner:
> And the yellow must take care that the the text is Black for accessibility

will take care of that with course card rework, right now it has shadows which makes it acceptable... and a major improvement, as before every color was possible

dominikbruechner:
screenshot 2018-12-12 at 18 33 32

Got the name back in real quick because of open PR and we have no more name on the platform right now

schurigh:
@dominikbruechner If it looks like on your screenshot, spacing should get fixed, looks weird, doesn't it?

dominikbruechner:
> @dominikbruechner If it looks like on your screenshot, spacing should get fixed, looks weird, doesn't it?

agreed, and optimised.
screenshot 2018-12-13 at 12 16 29

PR #962, opened 2018-12-17
remove school cap in login entirely

schurigh:
quickfix for https://ticketsystem.schul-cloud.org/browse/SC-661
I think we will never want to limit the school count in this dropdown so limit false should be okay here forever, right?


janrenz:
Mid term this should have some autocomplete lazy loading

PR #972, opened 2019-01-03
SC-656-Benachrichtigungseinstellungen ausblenden

KatharinaBlass:
ticket: https://ticketsystem.schul-cloud.org/browse/SC-656

### DIFF
--- a/views/account/settings.hbs
+++ b/views/account/settings.hbs
@@ -76,67 +76,67 @@
                 {{> "lib/forms/password-hint" plaintext=false style="display: inline;"}}
             </span>
         </div>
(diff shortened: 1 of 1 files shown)


views/account/settings.hbs
```diff
@@ -76,6 +76,7 @@
                 {{> "lib/forms/password-hint" plaintext=false style="display: inline;"}}
             </span>
         </div>
+<!-- the following is hidden until the feature is fixed
```
adrianjost: You should use `{{#if false}}` to comment out this section, cause HTML comments like here are still sender to the client but are hidden there and never used. This saves the user a couple of bytes and make him happy. 🙃

PR #974, opened 2019-01-07
[NC-76] csv import feedback verbessern

Parnswir:
Server PR: https://github.com/schul-cloud/schulcloud-server/pull/434

### DIFF
--- a/controllers/administration.js
+++ b/controllers/administration.js
@@ -434,10 +434,17 @@ const getCSVImportHandler = () => {
                     message: `${stats.users.successful} von ${numberOfUsers} Nutzer${numberOfUsers > 1 ? 'n' : ''} importiert.`,
                 };
             } else {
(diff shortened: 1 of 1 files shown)


schurigh:
image

controllers/administration.js
```diff
@@ -428,10 +428,25 @@ const getCSVImportHandler = () => {
                 },
             });
             const numberOfUsers = (stats.users.successful || 0) + (stats.users.failed || 0);
```
adrianjost: does this line work if no users where importet or do we need another fallback like `((stats.users || {}).súccessfull || 0) ...` ?

Parnswir: To be honest, I think I'll reduce it even further to `stats.users.successful + stats.users.failed`, as this is the default value with wich the statistics are initialized server-side.

PR #978, opened 2019-01-08
[WIP] implement resend pending invitations

schurigh:
https://ticketsystem.schul-cloud.org/browse/NC-62
still WIP, team service need to be modified by @CeEv

### DIFF
--- a/controllers/teams.js
+++ b/controllers/teams.js
@@ -785,7 +785,7 @@ router.get('/:teamId/members', async function(req, res, next) {
         ];
 
         const invitationActions = [{
-            class: 'btn-edit-invitation',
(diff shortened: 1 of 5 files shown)


controllers/teams.js
```diff
@@ -919,9 +920,23 @@ router.delete('/:teamId/members', async function(req, res, next) {
     res.sendStatus(200);
 });
 
+router.patch('/:teamId/invitation', async function(req, res, next) {
```
janrenz: why not move this to the server?

CeEv: It is only a proxy route. The logic is already handle from the api.

PR #980, opened 2019-01-10
defer execution of calculateMetrix to avoid missing values in window.…

tofixx:
let metrix calculation await all values are avalilable to avoid negative numbers

### DIFF
--- a/static/scripts/sw/metrix.js
+++ b/static/scripts/sw/metrix.js
@@ -100,6 +100,8 @@ window.addEventListener('load', function () {
       window.addEventListener('offline', updateOnlineStatus);
       updateOnlineStatus();
     }
-    calculateMetrics();
+    setTimeout(function () {
+      calculateMetrics();
+    }, 0); 
   }
 });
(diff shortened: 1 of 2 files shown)


tofixx:
@adrianjost  however travis is still failing, how to fix this?
 ```
Homework tests GET /homework/new:
     Error: Timeout of 15000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

adrianjost:
For me https://schul-cloud.org/homework/new isn't loading at all so there must be a bug and it's correct that the test is failing. But this has nothing Todo with your pull request. Will take a look on this later today!

adrianjost:
This thing broke with a server security update: `UnhandledPromiseRejectionWarning: StatusCodeError: 404 - {"name":"NotFound","message":"There are no lessons that you have access to."`

adrianjost:
The failing tests should be fixed here: https://github.com/schul-cloud/schulcloud-client/pull/982

PR #981, opened 2019-01-11
SC-699-Make-JWT-Cookie-available-to-subdomains

tofixx:
- by adding domain in cookie creation, the cookie will be available for all subdomains https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie
 - could not test the changes in registration.js locally
- await #982 to fix tests


tofixx:
@janrenz the cookie for explictly defined domain will not be set if it does not match the current domain name exactly. we can set it either for all subdomains or none of them (IE will always set cookies for all subdomains) 

the only way to set the jwt for the single api subdomain only is to call the api from clientside with the token given like https://stackoverflow.com/questions/3342140/cross-domain-cookies#answer-27232743

janrenz:
Ok, agree. Than we will move our tools out of global domain scope

MrSerth:
@janrenz Okay, I made a comment in the ticket. Is there anything else you would request to be changed? If not, could you please update your review?

PR #987, opened 2019-01-15
make jwt cookie available for subdomains

tofixx:
the cookie domain can be defined using env SC_DOMAIN, if given, the cookie will be available for subdomains as well. if not defined the domain parameter will not be set in cookie header definition


MrSerth:
Do we know what added the `www.` accidentally in our production deployment?

tofixx:
I would recommend to Change the value to be from configuration.

MrSerth:
Alright, if there is no other (dynamic) way, then we should do that. Otherwise, this PR is the same as #981.

controllers/registration.js
```diff
@@ -65,7 +65,7 @@ ${res.locals.theme.short_title}-Team`
         });
     }).then(function() {
         if (req.params.sso) {
-            res.cookie('jwt', req.cookies.jwt, {expires: new Date(Date.now() - 100000)});
+            res.cookie('jwt', req.cookies.jwt, { domain: req.hostname, expires: new Date(Date.now() - 100000) });
```
tofixx: we should update this to clearCookie instead

controllers/login.js
```diff
@@ -37,7 +37,7 @@ router.post('/login/', function (req, res, next) {
             } else {
                 const login = (data) => {
                     return api(req).post('/authentication', {json: data}).then(data => {
-                        res.cookie('jwt', data.accessToken, {expires: new Date(Date.now() + 30 * 24 * 60 * 60 * 1000)});
+                        res.cookie('jwt', data.accessToken, { domain: req.hostname, expires: new Date(Date.now() + 30 * 24 * 60 * 60 * 1000) });
```
tofixx: alternative strategy: use config value or ignore domain property

adrianjost:
@tofixx please resolve conflicts

tofixx:
Dank vieler Merges bereits im Master, getestet in FF und Chrome.

PR #991, opened 2019-01-17
[SC-505] - pwrecovery feedback notification

adrianjost:
image

image


package.json
```diff
@@ -46,6 +46,7 @@
     "moment": "^2.22.2",
     "morgan": "~1.7.0",
     "multer": "^1.3.0",
+    "natives": "^1.1.6",
```
Parnswir: What exactly is the reason for using this module? It [sounds unwise](https://www.npmjs.com/package/natives#caveat) to use it.

adrianjost:
Sorry für leaving this in. For any reason I can't start the schul-cloud with node ^10.15.0 on Windows 10 without this package installed. I wasn't able to figure out why but can reproduce it even with a clean install of the repo.
But there is no reason to leave this in here. I will remove it :)

PR #1005, opened 2019-01-23
Only load piwik in production environment

tim-friedrich:


janrenz:
We could also make this only if the piwik url and site is is set in env and make this dynamic.

janrenz:
@tim-friedrich ping...

janrenz:
Ping :)

adrianjost:
pong,
das sollten wir doch gemerged bekommen

PR #1008, opened 2019-01-25
Folder blacklist and misc

Langleu:
based on https://github.com/schul-cloud/schulcloud-server/pull/453
- [x] handle forbidden folders
- [x] add quick upload overview

image


Parnswir:
Wouldn't it be better to use an an icon instead of browser-specific unicode-characters for consistency? cc: @dominikbruechner

adrianjost:
Then we should be consistent and use icons instead of unicode charachters at the tasks section as well.
Or we could load _"our own"_ unicode-symbols font for consistency.
image

PR #1023, opened 2019-02-05
[SC-766] hide team administration

Parnswir:


Parnswir:
The feature isn't finished, so it would have to be disabled for the NBC as well 😄 I don't think it was planned to selectively enable the team administration page, but I would be happy to provide a fix if necessary @konstantin-sc

PR #1049, opened 2019-02-18
Turbolinks integration

tim-friedrich:
This change integrates turbolinks to a subset of tested routes. 
Turbolinks is needed to enable the webrtc based p2p cdn to prevent page loads.
It also reduces loading times.

To disable turbolinks just remove all routes from turbolinks.config.js


janrenz:
Will handle this in vue

PR #1050, opened 2019-02-18
[SC-914] fix consent emails being sent to random users

Parnswir:


controllers/administration.js
```diff
@@ -1164,31 +1164,35 @@ const getUsersWithoutConsent = async (req, roleName, classId) => {
...
-
-    const usersWithoutConsent = users.filter(user => {
-        let userWithConsent = consents.find(consent => {
-            return consent.userId === user._id;
-        });
+    const consents = (await api(req).get('/consents', {
+        qs: {
+            userId: { $in: users.map(u => u._id) },
+            $populate: 'userId',
+            $limit: false,
```
CeEv: Can you test if $limit=false send from client work? I expect that our global server hooks format the boolean to a string and therefore it do not work.

CeEv:
The name and logic in getUsersWithoutConsent sounds like a top level route for the api, that can call from client.

janrenz:
It sounds like this should be not a own Route in the server but maybe a filter

controllers/administration.js
Parnswir: **without $limit**

limit
-----------------------
**$limit: false**

limit_false

Parnswir:
Let's move the discussion of how and when to here: https://ticketsystem.schul-cloud.org/browse/SC-945

PR #1066, opened 2019-02-21
SC-952: Fix folder click in editor view

rodebert:


CeEv:
How is the test status, can we merge it? @Metauriel @kremer-io

PR #1075, opened 2019-02-27
SC-908: Fix buggy file move modal

rodebert:
Server-PR: https://github.com/schul-cloud/schulcloud-server/pull/498


Parnswir:
not sure if this is a client or server problem, but I can't move files into nested sub-directories (see screenshot: `jklö` is a sub-directory of `asdf` which is a directory in the files root)

asdf

rodebert:
@Parnswir Thanks for the hint. Should be fixed now.

PR #1082, opened 2019-03-04
SC-1015 debug logging

gisderdube:
fixes two bugs:

1. when deleting existing rss feeds, would open two modals
2. could not add new rss feeds for existing schools

Actual changes:
lines 2044 + 2111


gisderdube:
@Parnswir Couldn't squash manually, since the commit history was already pushed to remote. If desired, you have to `merge --squash`.

Note: I was trying to to `git reset HEAD~n --soft` for pre-merge squashing.

PR #1092, opened 2019-03-08
Adding data-attributes to login-test relevant elements

MxmRz:
Added data-testid attributes to login-test relevant elements. 
Please do not delete branch, further data-attributes will be added here. 
The data-testid attributes are currently intended to be used on the Schulcloud-testpage and need to be deployed there.


package.json
```diff
@@ -46,6 +46,7 @@
     "moment": "^2.22.2",
     "morgan": "~1.7.0",
     "multer": "^1.3.0",
+    "natives": "^1.1.6",
```
MrSerth: Why are you adding another package here?

MxmRz: I am not sure how this happened. I certainly did not add any package intentionally. 
I think it can be left out...

janrenz:
@MxmRz Ping ..

PR #1098, opened 2019-03-12
SC-942 fixed a bug where students could not submit group homework when max t…

gisderdube:
https://ticketsystem.schul-cloud.org/browse/SC-942
…eam size was not set. Also fixes a bug where students could sometimes not submit homework at all.


gisderdube:
backend pull is here: https://github.com/schul-cloud/schulcloud-server/pull/505

PR #1100, opened 2019-03-12
SC-871-add-students-from-previous-class

gisderdube:
https://ticketsystem.schul-cloud.org/browse/SC-871
could not reproduce the bug listed in the ticket. Instead fixed a bug where grades would not be sorted.


schurigh:
I couldn't reproduce the reported bug either so I fixed another thing that I noticed: Redirections. They were set to referrer, so redirect path could be annoying:

1. Klassenverwaltung (administration/classes) -> Klasse erstellen (/classes/create) -> Klasse managen (/classes/manage) -> Änderungen speichern --will redirect to--> Klasse erstellen (/classes/create)
2. Same if you go to Klasse verwalten (/manage) -> Klasse umbenennen (link below title, /edit) -> Änderungen speichern --will redirect back to--> Klasse verwalten (/manage).

PR #1108, opened 2019-03-14
Editor

CeEv:


buildnpush.sh
```diff
@@ -0,0 +1,10 @@
+#! /bin/bash
```
adrianjost: What is this file used for? I can't find any reference to it.
If it is for deployment or something similar, I would recommend adding it to the package.json with the `cross-env` package. Then it can run on windows as well as on OSX & UNIX.

```json
{
  "scripts": {
    "deploy": "cross-env sh ./buildnpush.sh"
  }
}
```

I would also recommend renaming this file to something without shortend words like `and` -> `n`. => `buildAndPush.sh`?

PR #1136, opened 2019-04-03
check if birthday is set

CordlessWool:
[SC-1144](https://ticketsystem.schul-cloud.org/browse/SC-1144)


controllers/administration.js
```diff
@@ -984,8 +984,11 @@ router.get('/teachers/:id/edit', permissionsHelper.permissionsChecker(['ADMIN_VI
 
 const getStudentUpdateHandler = () => {
 	return async function (req, res, next) {
-		const birthday = req.body.birthday.split('.');
-		req.body.birthday = `${birthday[2]}-${birthday[1]}-${birthday[0]}T00:00:00Z`;
+
+		if (req.body.birthday) {
+			let birthday = req.body.birthday.split('.');
```
adrianjost:
```suggestion
			const birthday = req.body.birthday.split('.');
```

PR #1268, opened 2019-06-28
set keep alive header for external and api requests

tofixx:
requires KEEP_ALIVE in boss client


FalkoBenthin:
KEEP_ALIVE in Boss-Cloud-Client gesetzt

PR #1288, opened 2019-07-05
SC-1595 - optimize gulp speed (on ci)

adrianjost:
## Inhalt
- optimizations:
  - only run mandatory scripts on ci (skip images, ...)
  - use symlinks for static files instead of copying those into the build folder
- gulpfile refactoring
  - executed eslint
  - added some code comments 
  - ordered imports alphabeticly
- **~2min faster** than before

## Ticket
- [x] Link zum Ticket https://ticketsystem.schul-cloud.org/browse/SC-1595


gulpfile.js
```diff
@@ -201,22 +213,45 @@ gulp.task('vendor-scripts', () => beginPipe('./static/vendor/**/*.js')
...
 	.pipe(gulp.dest(`./build/${themeName()}/vendor`)));
 
+
+// =========================
+// COPY (Symlink) Files
+// =========================
+
 // copy other vendor files
-gulp.task('vendor-assets', () => beginPipe([
+gulp.task('vendor-assets', () => beginPipeLink([
```
tofixx: is this linking all file or just the folders which should be enough

adrianjost: Good point. It currently links all files individually. I will adjust this :)

adrianjost: Okay, I can't figure out how to link the whole folder at once. As far as I understand gulp now, it only reads files into the pipeline. Therefore we can't symlink folders, because they simply can't get read into the task.

gulpfile.js
```diff
@@ -201,22 +213,45 @@ gulp.task('vendor-scripts', () => beginPipe('./static/vendor/**/*.js')
...
 	'!./static/vendor/**/*.{css,sass,scss}',
 ]).pipe(gulp.dest(`./build/${themeName()}/vendor`)));
 
 // copy vendor-optimized files
-gulp.task('vendor-optimized-assets', () => beginPipe(['./static/vendor-optimized/**/*.*'])
+gulp.task('vendor-optimized-assets', () => beginPipeLink(['./static/vendor-optimized/**/*.*'])
 	.pipe(gulp.dest(`./build/${themeName()}/vendor-optimized`)));
 
 // copy node modules
 const nodeModules = ['mathjax', 'font-awesome'];
```
tofixx: here we should add information to node modules, which files have to be copied only. if nodeModules gets objects with name and regex's that could work. most node modules have only a build folder that is used

adrianjost: done

Parnswir:
We need to wait for the release to be merged back or cherry-pick bdbfcee82943c3578d446f3f082ffed8fa30a848 and e5ed38a729b56eb26fc6d0af00bcf016da90fcab (from the release branch) to make the tests pass. 😭

gulpfile.js
```diff
@@ -201,22 +213,46 @@ gulp.task('vendor-scripts', () => beginPipe('./static/vendor/**/*.js')
...
+gulp.task('vendor-optimized-assets', () => beginPipeLink(['./static/vendor-optimized/**/*.*'])
 	.pipe(gulp.dest(`./build/${themeName()}/vendor-optimized`)));
 
 // copy node modules
-const nodeModules = ['mathjax', 'font-awesome'];
-gulp.task('node-modules', () => Promise.all(nodeModules
-	.map(module => beginPipe([`./node_modules/${module}/**/*.*`])
-		.pipe(gulp.dest(`./build/${themeName()}/vendor-optimized/${module}`)))));
+const nodeModules = [
+	['mathjax', 'mathjax/MathJax.js'],
```
tofixx: can u create a single vendor js and css file for all vendor-optimized snippets instead of symlinking?

adrianjost: you mean like we did with sentry?

tofixx: yop, using that we would no more have all the mathjax files to be copied here file by file

adrianjost:
not very usefull

PR #1312, opened 2019-07-29
SC-1645 Cors middleware

tofixx:
[x] to enable CORS and Security Header set the environment variables CORS and SECURITY_HEADERS to 1 @FalkoBenthin


FalkoBenthin:
was soll denn nun wohin gemountet werden?

malcm:
@kremer-io should I add all known URLS to the /content route?

app.js
```diff
@@ -4,16 +4,35 @@ const logger = require('morgan');
...
+
+// template stuff
+const authHelper = require('./helpers/authentication');
+
+// mung middleware
+const securityHeaders = require('./middleware/security_headers');
+const cors = require('./middleware/cors');
+
+// set custom response header for ha proxy
+if (process.env.KEEP_ALIVE) {
```
tofixx: move down, to habe app.use of requires above directly where it is used

middleware/cors.js
```diff
@@ -0,0 +1,75 @@
+const cspConfig = require('../config/http-headers').content_security_policy;
```
tofixx: use object destructoring and CamelCase instead

middleware/cors.js
```diff
@@ -0,0 +1,75 @@
...
+			objectSrc = `${objectSrc} ${matchingHeader.objectSrc}`;
+		}
+	});
+	const finalCors = { defaultSrc, scriptSrc, objectSrc };
+	logger.debug(`cors headers for route ${path}`, finalCors);
+	return finalCors;
+};
+
+const accessControlHeadersForRoute = (path, regexs) => {
+	const matchingKeys = Object.keys(regexs)
```
tofixx: this duplication should be resolved

middleware/security_headers.js
```diff
@@ -0,0 +1,18 @@
...
+
+if (!process.env.SECURITY_HEADERS) 	{
+	logger.info('SECURITY_HEADERS env has not been defined, to enable'
+	+ ' security header set value to 1 and update in config/http-headers.js');
+}
+
+const cors = (req, res, next) => {
+	if (process.env.SECURITY_HEADERS) {
+		Object.keys(securityHeaderConfig).forEach((header) => {
+			res.setHeader(header, securityHeaderConfig[header]);
```
tofixx: securityHEADERconfig should only include header :)

package.json
```diff
@@ -21,6 +21,7 @@
     "body-parser": "^1.19.0",
     "browser-sync": "^2.26.7",
     "compression": "^1.6.2",
+    "config": "^3.2.2",
```
tofixx: remove

middleware/cors.js
malcm: the logic is the same except for 3 lines with cspHeadersForRoute. Makes sense to have a new function for it.

middleware/security_headers.js
malcm: setHeader just work with 2 parameter. So makes sense like this. Otherwise we need to start to split the string ...

middleware/cors.js
malcm: ok

package.json
malcm: ok

tofixx:
> @kremer-io should I add all known URLS to the /content route?

@malcm @kremer-io @FalkoBenthin 
please discuss, how to have the correct values be defined in the live system, for developement reasons and test systems we will not need to add more details inside of the predefined configuration. Please cover the use case that we have different systems running which need different configurations and they should easily be updated

app.js
malcm: Done

app.js
```diff
@@ -4,16 +4,35 @@ const logger = require('morgan');
 const cookieParser = require('cookie-parser');
 const bodyParser = require('body-parser');
 const compression = require('compression');
-
 const session = require('express-session');
-
-// template stuff
+const methodOverride = require('method-override');
```
tofixx: is that used?

malcm: I don't know. But you just changed the position of the import. I will keep it for now

malcm:
> 
> 
> > @kremer-io should I add all known URLS to the /content route?
> 
> @malcm @kremer-io @FalkoBenthin
> please discuss, how to have the correct values be defined in the live system, for developement reasons and test systems we will not need to add more details inside of the predefined configuration. Please cover the use case that we have different systems running which need different configurations and they should easily be updated

@FalkoBenthin we will have a look on monday how we can deploy it

malcm:
@schul-cloud/client-core please review for merge
@FalkoBenthin is aware of deployment

FalkoBenthin:
jinja2-Template für http_headers.js geschnürt und Ansibles Templates angepasst, CORS und SECURITY_HEADERS testweise für demo auf 1 gesetzt

PR #1324, opened 2019-08-01
Sc 1599 min video player

adrianjost:
replaced the OpenHPI video player with the native HTML5 player. This reduces the loaded overhead and the browser can decide which data it will load (`preload="auto"`).

## Allgemein
- [x] Link zum Ticket https://ticketsystem.schul-cloud.org/browse/SC-1599


adrianjost:
@tofixx @schurigh ping

malcm:
@adrianjost good point but keep in mind native HTML5 player does not have HLS support in all browsers. If needed in future

adrianjost:
@malcm true, but it was common sense that we don't need it for our simple use cases.

malcm:
@adrianjost ok good. I just remember that @janrenz wants to have it in the schul-cloud

adrianjost:
@malcm When I discussed this with him the last time (a month ago) he liked the idea of removing it for performance reasons 😅

malcm:
👍

static/styles/homepage/homepage.scss
```diff
@@ -27,3 +27,7 @@ h1.section-title{
         font-size: 36px;
     }
 }
+
+video{
```
tofixx: lint

static/styles/lib/base.scss
```diff
@@ -89,6 +89,9 @@ p.small {
 label {
 	font-weight: bold;
 }
+video, audio {
+	width:100%;
```
tofixx: lint

tofixx:
@adrianjost please fix the height issue and resolve conflicts that we may include this pr in the next release asap

adrianjost:
@tofixx adjusted it and should work in all modern browsers (https://caniuse.com/#search=object-fit)

PR #1350, opened 2019-08-13
add sentry error reporting for client application

tofixx:
# Checkliste

requires SENTRY_DSN to be defined: `https://1196aaf71f2e4f38b83afa3db1c30ecc@sentry.schul-cloud.dev/3` can be used for testing reasons locally

- [x] some error are not reported and sentry fails that should be fixed
  - [x] video.js
  - [x] not authorized
- [x] client node app reporting needs to be added
- [x] data security review required: userId will be logged @schurigh 
- - [x] remove IDs from URI
- - [x] make sure that no requester IP address is transmitted and/or stored
- [x] the required tags and env values have to be defined, see sentry init and scope for details @Parnswir


views/lib/default.hbs
```diff
@@ -2,6 +2,7 @@
 <html lang="de">
 
 <head>
+{{#embed "lib/sentry"}}{{/embed}}
```
adrianjost: indentation

views/lib/sentry.hbs
```diff
@@ -0,0 +1,34 @@
+{{#ifneq @root.SENTRY_DSN false}}
+<script src="/vendor-optimized/@sentry/browser/build/bundle.min.js"></script>
+<script src="/vendor-optimized/@sentry/integrations/build/captureconsole.js"></script>
+<script>
```
adrianjost: Maybe you can move all this code into a .js file and import the two sentry modules there and initialize it. Than we wouldn't need to copy the complete package with gulp and we could maybe load the script async. The way it is now, it blocks the rendering process and therefore increases the time to first paint.

gulpfile.js
```diff
@@ -213,7 +213,7 @@ gulp.task('vendor-optimized-assets', () => beginPipe(['./static/vendor-optimized
 	.pipe(gulp.dest(`./build/${themeName()}/vendor-optimized`)));
 
 // copy node modules
-const nodeModules = ['mathjax', 'font-awesome'];
+const nodeModules = ['mathjax', 'font-awesome', '@sentry'];
```
adrianjost: you don't need this if my idea below works

views/lib/sentry.hbs
tofixx: however there are variables to be replaced which will not work in plain js files :/ otherwise i really like that idea!

adrianjost: Ahh haven't seen this.

PR #1375, opened 2019-09-02
auto generate currentYear

adrianjost:
# Checkliste

## Allgemein
- [x] Link zum Ticket https://ticketsystem.schul-cloud.org/browse/SC-1933

## Mehr

### DIFF
--- a/theme/n21/views/lib/footer.hbs
+++ b/theme/n21/views/lib/footer.hbs
@@ -1,7 +1,7 @@
 <footer class="hide-inline small-footer">
 	<div class="imprint">
 		<div class="center">
-			<a>&copy; 2018 </a>
+			<a>&copy; {{currentYear}} </a>
 
 			<div>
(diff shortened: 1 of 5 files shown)


theme/n21/views/lib/footer.hbs
```diff
@@ -1,7 +1,7 @@
 <footer class="hide-inline small-footer">
 	<div class="imprint">
 		<div class="center">
-			<a>&copy; 2018 </a>
+			<a>&copy; {{currentYear}} </a>
```
tofixx: is the theme short title removed here intentionally?

adrianjost:
@tofixx lint issues should be solved now.
empty line at the end of all edited files was already there

theme/n21/views/lib/footer.hbs
adrianjost: I don't know, but that's not part of this ticket.

PR #1376, opened 2019-09-02
SC-1837 schuljahr bearbeiten bugfix (handlebars nesting)

adrianjost:
# Checkliste

## Allgemein
- [x] Link zum Ticket https://ticketsystem.schul-cloud.org/browse/SC-1837

## Mehr


adrianjost:
with root we are safe and shouldn't need to adopt anything here if we ever update to handlebars 4

PR #1378, opened 2019-09-02
SC-1881 - removed some student settings

adrianjost:
# Checkliste

## Allgemein
- [x] Link zum Ticket https://ticketsystem.schul-cloud.org/browse/SC-1881

## Mehr


helpers/handlebars/helpers/index.js
```diff
@@ -133,6 +133,9 @@ module.exports = {
 		}
 		return opts.inverse(this);
 	},
+	userHasRole: (role, opts) => {
+		return !!opts.data.local.currentUser.roles.find((r) => r.name === role);
```
tofixx: check that logic is nowhere else used in the client and replace it using that helper ... is that function relly nowhere else already defined?

adrianjost: @tofixx updated it

dominikbruechner: great work @adrianjost 

love that all this **** is gone now :)

PR #1424, opened 2019-09-30
Feature/test sentry for client

tofixx:
# Checkliste
small addition to #1350


controllers/login.js
```diff
@@ -17,6 +17,9 @@ const getSelectOptions = (req, service, query) => api(req).get(`/${service}`, {
 }).then(data => data.data);
 
 const clearCookie = (req, res) => {
+	if (req.session && req.session.destroy) {
```
Parnswir: How is this set?

adrianjost: I think it's more some kind of check, if the method exists and can be executed.
Maybe it's needed if the session should get deleted but isn't yet created for the current user?
@tofixx

PR #1426, opened 2019-10-01
SC-1315 Do not return API errors to the user

Parnswir:
see title

## Allgemein
- [X] Link zum Ticket https://ticketsystem.schul-cloud.org/browse/SC-1315


janrenz:
Should we display a generic error?

adrianjost:
> Should we display a generic error?

We already do. Our default error page wich is now used has a generic error message.

PR #1497, opened 2019-11-18
disable serviceworker due to different errors and csrf token caching

tofixx:
# Checkliste
this disables the current registered serviceworkers which has to be unregistered due to caching side effects. a new reimplementation will be done in the nuxt client


tofixx:
merged into different hotfix

PR #1525, opened 2019-12-02
Feature/tsp visibility sc 2173

tofixx:
# Checkliste

## Allgemein
- [x] Links zu verwandten PRs anderer Repositories
  - https://github.com/schul-cloud/schulcloud-server/pull/887
  - Im Ticket (oder PR, wenn kein Ticket): Beschreibung und Begründung der Änderung/Neuerungen
- [x] Link zum Ticket https://ticketsystem.schul-cloud.org/browse/SC-2173

### DIFF
--- a/app.js
+++ b/app.js
@@ -13,13 +13,18 @@ const handlebars = require('handlebars');
 const layouts = require('handlebars-layouts');
 const handlebarsWax = require('handlebars-wax');
 const Sentry = require('@sentry/node');
+const { Configuration } = require('@schul-cloud/commons');
+
 const { tokenInjector, duplicateTokenHandler } = require('./helpers/csrf');
 
 const { version } = require('./package.json');
 const { sha } = require('./helpers/version');
 const logger = require('./helpers/logger');
 
+
 const app = express();
+const config = new Configuration({ schemaFileName: 'default.schema.json' });
+config.init(app);
 
 if (process.env.SENTRY_DSN) {
 	Sentry.init({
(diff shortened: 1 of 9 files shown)


config/default.schema.json
```diff
@@ -0,0 +1,20 @@
...
+		"TEACHER_VISIBILITY_FOR_EXTERNAL_TEAM_INVITATION": {
+			"type": "string",
+			"default": "opt-out",
+			"enum": [
+				"disabled",
+				"opt-in",
+				"opt-out",
+				"enabled"
+			],
+			"description": "defines wheter external team invitation shows teachers from different schools or not. if enabled system wide there are options general enabled or opt-in/-out by user required."
```
Parnswir: See my suggestion in the server PR

controllers/account.js
```diff
@@ -16,24 +16,28 @@ router.post('/', (req, res) => {
...
 		passwordNew,
 	} = req.body;
-	const discoverable = !!req.body.discoverable;
+	let discoverable = null;
+	if (req.body.discoverable === 'true') {
+		discoverable = true;
+	}
+	if (req.body.discoverable === 'false') {
+		discoverable = false;
+	}
```
adrianjost: shouldn't `discoverable` default to `false`? `null´ is handled like `false` in our code so it is unnecessary to have the second if.

app.js
```diff
@@ -13,13 +13,18 @@ const handlebars = require('handlebars');
 const layouts = require('handlebars-layouts');
 const handlebarsWax = require('handlebars-wax');
 const Sentry = require('@sentry/node');
+const { Configuration } = require('@schul-cloud/commons');
```
adrianjost: `Configuration` is really generic and difficult to understand. I have no clue what this package is doing.
After reading the npm package description I understand what it should do, but I still think the name should be changed here.
There is also no documentation at all, how to use and configure that package.

controllers/account.js
tofixx: nope, we have a nullable boolean here so it has three states, not only two.

PR #1527, opened 2019-12-04
MDG Nienburg entfernt

JacquelineThorns:
# Checkliste


janrenz:
we can also delete the logo from the S3 bucket

JacquelineThorns:
@janrenz done

PR #1540, opened 2019-12-11
remove nodemon from dockerfile

adrianjost:
nodemon is not used in production, so there is no need for it. It may be the reason why the build is failing from time to time

https://github.com/npm/uid-number/issues/3
https://github.com/npm/npm/issues/20861
https://jira.atlassian.com/browse/BCLOUD-16334


tofixx:
already cherrypicked

PR #1618, opened 2020-02-19
Update dependencies

Parnswir:
More focussed version of #1555


gulpfile.js
```diff
@@ -297,23 +294,6 @@ gulp.task('watch', ['build-all'], () => {
 	gulp.watch(withTheme('./static/*.*'), watchOptions, ['static']);
 });
 
-gulp.task('watch-reload', ['watch', 'browser-sync']);
```
adrianjost: would be nice to keep this. I use it every time

adrianjost: sorry, just overlooked it.

PR #1629, opened 2020-03-04
create npm audit workflow

adrianjost:
# Checkliste

- successor of https://github.com/schul-cloud/schulcloud-client/pull/1557

## Allgemein
- [x] Link zum Ticket https://ticketsystem.schul-cloud.org/browse/OPS-247

## Mehr


Parnswir:
Nice, but it fails due to dev-dependencies right now although it shouldn't with `--only=prod` 😄

adrianjost:
@Parnswir It's a known bug 😅🙄

https://npm.community/t/npm-audit-without-fix-ignores-only-prod/3959/7

with an open PR

https://github.com/npm/cli/pull/125

adrianjost:
I have now seperated DEV & PROD checks so we can monitor them seperatly as long as they are not both green

PR #1631, opened 2020-03-06
ce/ stable and performance fix for show course

CeEv:
# Checkliste

## Mehr


controllers/courses.js
```diff
@@ -12,6 +12,8 @@ const logger = require('../helpers/logger');
 
 const router = express.Router();
 
+const { CALENDAR_SERVICE_ENABLED, HOST } = process.env;
```
tofixx: This is a feature, we use `FEATURE_...` named ENV's in other situations, why not here?

controllers/courses.js
```diff
@@ -620,99 +622,122 @@ router.get('/:courseId/', (req, res, next) => {
 	];
 
 	// ########################### start requests to new Editor #########################
+	let editorBackendIsAlive = true;
 	if (EDITOR_URL) {
```
tofixx: This sounds like a feature, why translating the existence of an url into a feature flag?
Instead I would use a feature flag and if enabled, the url (and maybe other values as well) should be defined.

controllers/courses.js
```diff
@@ -620,99 +622,122 @@ router.get('/:courseId/', (req, res, next) => {
...
-		promises.push(apiEditor(req).get(`course/${req.params.courseId}/lessons`));
+		promises.push(apiEditor(req)
+			.get(`course/${req.params.courseId}/lessons`)
+			.catch((err) => {
+				logger.warn('Can not fetch new editor lessons.', err);
```
tofixx: this is an error, isn't it?
```suggestion
				logger.error('Can not fetch new editor lessons.', err);
```

controllers/courses.js
```diff
@@ -620,99 +622,122 @@ router.get('/:courseId/', (req, res, next) => {
...
+					total: 0,
+					data: [],
+				};
+			}));
 	}
 
 	// ############################ end requests to new Editor ##########################
+	const [
+		course,
+		_lessons,
```
tofixx: lint: no underscore?

controllers/courses.js
```diff
@@ -620,99 +622,122 @@ router.get('/:courseId/', (req, res, next) => {
...
-				url: `/courses/${req.params.courseId}/topics/${lesson._id}/`,
-			}));
-
-			const homeworks = (_homeworks.data || []).map((assignment) => {
-				assignment.url = `/homework/${assignment._id}`;
-				return assignment;
-			});
+	try {
+		// ############################## check if new Editor options should show #################
+		const edtrUser = (res.locals.currentUser.features || []).includes('edtr');
```
tofixx: use more descriptive name like `currentUserHasFeatureEditor`

controllers/courses.js
```diff
@@ -620,99 +622,122 @@ router.get('/:courseId/', (req, res, next) => {
...
+		const userHasEditorEnabled = EDITOR_URL && edtrUser;
```
tofixx: the meaning is not correct? why doing this if feature is globally disabled?

controllers/courses.js
```diff
@@ -620,99 +622,122 @@ router.get('/:courseId/', (req, res, next) => {
...
-					),
-					// #################### new Editor, till replacing old one ######################
-					newLessons,
-					isNewEdtrioActivated,
-					scopedCoursePermission: scopedPermissions[res.locals.currentUser._id],
-				}),
-			);
-		})
-		.catch(next);
+		const baseUrl = req.headers.origin || HOST || 'http://localhost:3100';
```
tofixx: move this (`HOST || 'http://localhost:3100'`) to config/globals.js or use HOST only

controllers/courses.js
CeEv: I do not add this envs, only remove it from request callback into global scope.

controllers/courses.js
CeEv: The same i do not add it. It is the same like host and server urls. This is not a feature. The new editor is a base MS in future. But the client should work without, like calendar and content MS.

controllers/courses.js
CeEv: I do not add it and i do not know for what origin in context of CSRF is used. It is adding by aron.

controllers/courses.js
CeEv: It is old code too. In this PR i add only editorBackendIsAlive check and resort the error handling.

controllers/courses.js
CeEv: currentUser is the send user information from server and this includes features as flag. The model has the flag features.

controllers/courses.js
CeEv: Some closed beta tester can have the flag to use the new editor. If the editor backend is set and alive and a visible editor lesson exist, it should display it to the students.

PR #1648, opened 2020-03-12
22.3.0 SC-3586

tofixx:
# Checkliste

handle link redirection inside the client & handle errors in client
https://github.com/schul-cloud/schulcloud-server/pull/1035

PR #1649, opened 2020-03-12
Hotfix/22.3.1

adrianjost:
# #1648 should be merged first!

Ticket: https://ticketsystem.schul-cloud.org/browse/SC-3544

if the original form request contained a `_csrf` field, the token was also appended to the CSRF retry form. This resulted in multiple csrf tokens. I added some better error handling and prevented duplicate csrf tokens on the retry page.

### DIFF
--- a/helpers/csrf.js
+++ b/helpers/csrf.js
@@ -1,4 +1,5 @@
-/* eslint-disable no-underscore-dangle, max-len */
+/* eslint-disable no-underscore-dangle */
+const logger = require('./logger');
 
 const tokenInjector = (req, res, next) => {
 	res.locals.csrfToken = req.csrfToken();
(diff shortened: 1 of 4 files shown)


Parnswir:
Two questions:

- could the problem also be solved by taking the last token instead of the first token from the list?
- Why throw a warning in case multiple identical tokens exist? Nothing bad happens and no other request or process is affected by this

adrianjost:
@tofixx 

> * could the problem also be solved by taking the last token instead of the first token from the list?

not reliably, because it depends on the order of interpretation of node and in which order the browser do send the values

> * Why throw a warning in case multiple identical tokens exist? Nothing bad happens and no other request or process is affected by this

Because it is not intended and supported behaviour to have multiple CSRF Tokens. Otherwise we could start adding a thousand tokens and try all of them until we have a valid one.
I throw a warning, because we should get noticed if we implemented something wrong anywhere and this behaviour happens. Who knows if the tokens will be always in sync? Better be safe and get informed if something happens we do not expect and support so we can start fixing the issue before something bad happens.

malcm:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/schul-cloud/schulcloud-client/pullRequest?prid=5168705&bid=16967920)

Parnswir:
Good points, :+1:

PR #1659, opened 2020-03-16
Hotfeature/gulp clear cache keep build

tofixx:
# Checkliste

added `gulp clear-cache` which allows to clear the cache of gulp and webpack only without removal of build files. after `gulp clear-cache`, `gulp` is completely recreating the build files. 

if a build not consists most current files they can be updated without removal now.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -255,6 +255,19 @@ gulp.task('clear', () => gulp
 	)
 	.pipe(rimraf()));
 
+// clear gulp cache without removing current build	
+gulp.task('clear-cache', () => gulp
+	.src(
+		[
+			'./.gulp-changed-smart.json',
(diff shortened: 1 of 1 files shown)


tofixx:
> What's the benefit of this?
> Removing all files shouldn't take that much longer and as you mentioned all files still need to be generated again.

currently if we detect malicious data, we have to remove all build files and then can rebuild them, during this time, the client will not be usable while if we just override them, the usability is almost not affected

adrianjost:
> > What's the benefit of this?
> > Removing all files shouldn't take that much longer and as you mentioned all files still need to be generated again.
> 
> currently if we detect malicious data, we have to remove all build files and then can rebuild them, during this time, the client will not be usable while if we just override them, the usability is almost not affected

Ahh so this fix is for fast production rebuilds, right?

PR #1662, opened 2020-03-16
Feature/sc 3560 text to var

Tratori:
- [x] Link zum Ticket https://ticketsystem.schul-cloud.org/browse/SC-3560


locales/de.json
```diff
@@ -19,6 +34,60 @@
 			"startM2018": "(Start Mai 2018)"
 		}
 	},
+	"courses": {
+		"course": {
```
adrianjost: I have updated the guidelines to prefix dynamic url parts (id's) with an underscore. That should make it easier for us to identify those and not get crazy because we can not find `/courses/course`.
I will update this in your PR, because it was my fault to have forgotten this part in the documentation.
image

adrianjost:
Travis - Mocha läuft nur nicht, weil content-server gerade nicht erreichbar ist. Ansonsten sieht es gut aus.

PR #1745, opened 2020-04-08
Fix: adjust eslint config to match CI

maxklenk:
My IDE and `npm run lint` suggest to: Arrow functions can omit parentheses when they have exactly one parameter.
Screenshot 2020-04-08 at 12 36 04 PM

But on Codacy an error is reported when the are removed. My changes align the eslint config with the codacy handling.

eslint rule:
https://eslint.org/docs/rules/arrow-parens#requireforblockbody

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,13 +5,6 @@
 		"plugin:json/recommended"
 	],
 	"rules": {
-		"arrow-parens": [
-			"error",
-			"always",
-			{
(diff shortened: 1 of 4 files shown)


tofixx:
I just updated airbnb, if I remember correctly this rule has been changed in the last update which makes your manual changes unnecessary. Probably if so, we can add a linter update in this commit on the whole code / or better right afterwards in a further commit.

PR #1769, opened 2020-04-21
SC-4260 (former Hotfix/22.9.8) enable sentry sample rate

tofixx:
# Checkliste

move sentry settings to configuration and reduce reported number of issues, this replaces #1768


views/lib/sentry.hbs
```diff
@@ -22,7 +30,7 @@
 	});
 	Sentry.configureScope((scope) => {
 		scope.setLevel('warn');
-		{{#if @root.currentUser}}
+		{{#if @root.currentUser }}
```
adrianjost, line 33:
```suggestion
		{{#if @root.currentUser}}
```

PR #1770, opened 2020-04-21
feature / sc-3560-topic

abralurrahman:
- [x] Link zum Ticket https://ticketsystem.schul-cloud.org/browse/SC-3560


locales/de.json
```diff
@@ -718,5 +718,44 @@
 			"released": "Veröffentlicht",
 			"unpublished": "Unveröffentlicht"
 		}
+	},
```
CeEv, line 721: The same can you please add this to en.json too? :)

PR #1867, opened 2020-05-20
Fixed video preview in files

janrenz:
While testing videos loading I noticed that the video was not played. This fix made the video play again.


BlackGhostHD:
These changes are likely to relate to: https://ticketsystem.schul-cloud.org/browse/SC-4933
@ABialoskorski can you test if that will solve the ticket?

ABialoskorski:
I have a problem on localhost to upload a video from drive, I created ticket for that : https://ticketsystem.schul-cloud.org/browse/SC-4959.
But I can add a video from url,

PR #1898, opened 2020-06-04
OPS-456 move cookie secret to environment var

pkleybolte:
# Description

https://ticketsystem.schul-cloud.org/browse/OPS-456

## Approval for review


config/default.schema.json
```diff
@@ -274,6 +274,10 @@
 				"EXPIRES_SECONDS"
 			]
 		},
+		"COOKIE_SECRET": {
+			"type": "string",
+			"description": "Secret used to sign the session ID cookie"
+		},
```
Parnswir, line 280: to not break anything (and according to Jan's comment in the ticket), this should default to 'secret'

config/default.json
```diff
@@ -13,5 +13,6 @@
 		"SECURE": false,
 		"EXPIRES_SECONDS": 2592000000
 	},
+	"COOKIE_SECRET": "ihasDF(=4jA4aF",
```
Parnswir, line 16: Is this necessary when using the Configuration?

config/default.schema.json
pkleybolte: I talked to Jan. He agreed that it is safer to use a new password. Maybe some old sessions might invalidate, but that's not a problem.

config/default.json
pkleybolte: Could be moved to default.schema.json. We agreed to leave it here to "force" to set it explicitly in the instances and not use the default, that is publicly readable by everybody cloning the repo.

PR #1910, opened 2020-06-08
SC-4613 - registration-change-text-in-consent-screen

ABialoskorski:
# Description

- https://ticketsystem.schul-cloud.org/browse/SC-4613

### DIFF
--- a/locales/de.json
+++ b/locales/de.json
@@ -1157,7 +1157,8 @@
 					"password": "Passwort: ",
 					"theEmailAddressEnables": "Die E-Mail-Adresse ermöglicht den Versand von Benachrichtigungen und erleichtert das Zurücksetzen des Passwortes.",
 					"weNeedFollowingDataForRegistration": "Für die Nutzung der {{shortTitle}} benötigen wir folgende Angaben:",
-					"weStoreUserDataSafely": "Die {{shortTitle}} bewahrt die Nutzerdaten sicher auf und gibt sie nicht an Dritte weiter. Die Verarbeitung der Daten erfolgt entsprechend der hohen gesetzlichen Datenschutz-Anforderungen."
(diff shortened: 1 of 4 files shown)


locales/de.json
```diff
@@ -1159,7 +1159,8 @@
 					"password": "Passwort: ",
 					"theEmailAddressEnables": "Die E-Mail-Adresse ermöglicht den Versand von Benachrichtigungen und erleichtert das Zurücksetzen des Passwortes.",
 					"weNeedFollowingDataForRegistration": "Für die Nutzung der {{shortTitle}} benötigen wir folgende Angaben:",
-					"weStoreUserDataSafely": "Die {{shortTitle}} bewahrt die Nutzerdaten sicher auf und gibt sie nicht an Dritte weiter. Die Verarbeitung der Daten erfolgt entsprechend der hohen gesetzlichen Datenschutz-Anforderungen."
+					"weStoreUserDataSafelyStudent": "Die {{shortTitle}} bewahrt die Nutzerdaten sicher auf und gibt sie nicht an Dritte weiter. Die Verarbeitung der Daten erfolgt entsprechend der hohen gesetzlichen Datenschutz-Anforderungen.",
+					"weStoreUserDataSafely": "Für die Nutzung der {{shortTitle}} bedarf es für Schüler unter {{age}} Jahren der Einwilligung eines Erziehungsberechtigten"
```
BlackGhostHD, line 1163: Key and value don't match

BlackGhostHD, line 1163: weStoreUserDataSafely - key and value don't match

ABialoskorski:
> weStoreUserDataSafely is also used in registration-employee.hbs. The text there should probably not change.

ok, reverted

locales/de.json
ABialoskorski: what does it mean? I changed weStoreUserDataSafelyStudent back to weStoreUserDataSafely, and new one I named weStoreUserDataSafelyParent

BlackGhostHD: weStoreUserDataSafely has nothing to do with the content. the key describes something, which is then not available in terms of content.

BlackGhostHD: this would now apply to weStoreUserDataSafelyParent

PR #2005, opened 2020-07-15
[SC-5643] Dashboard A11y

pawelchoniawko:
# Description
Dashboard Accessibility errors fixing

https://ticketsystem.schul-cloud.org/browse/SC-5643


views/dashboard/components/homework_tiles.hbs
```diff
@@ -8,7 +8,7 @@
 	{{/if}}
 	<div data-testid="courses">
 		{{#each content}}
-			<a class="homework-link" href="{{this.url}}{{#if ../redirectToTab}}#activetabid={{../redirectToTab}}{{/if}}">
+			<a class="homework-link" aria-label="{{this.courseId.name}}" href="{{this.url}}{{#if ../redirectToTab}}#activetabid={{../redirectToTab}}{{/if}}">
```
dominikbruechner, line 11: should be course name + home work title I'd suggest

dominikbruechner:
should we have a look on the news card a11y here as well? maybe simply adding an aria label, such as with the assignments

haniapalac:
I noticed a weird tabbing order for the time-table on staging
image
first el -> third el -> second el. @pawelchoniawko please make double sure it works properly (at the moment I cannot check it locally due to server errors)

luizzaly:
Bildschirmfoto 2020-07-22 um 16 32 01

Should the title of the news card be focusable, because at the moment you jump directly to "weiterlesen" without knowing what the news is about... what do you think?

views/dashboard/components/homework_tiles.hbs
```diff
@@ -8,7 +8,7 @@
 	{{/if}}
 	<div data-testid="courses">
 		{{#each content}}
-			<a class="homework-link" href="{{this.url}}{{#if ../redirectToTab}}#activetabid={{../redirectToTab}}{{/if}}">
+			<a class="homework-link" aria-label="{{this.courseId.name}} - {{this.title}}" href="{{this.url}}{{#if ../redirectToTab}}#activetabid={{../redirectToTab}}{{/if}}">
```
haniapalac, line 11: I'm not 100% sure if it applies here but this could be the case where the user should be informed about the redirect or opening a new tab: https://www.w3.org/TR/WCAG20-TECHS/G201.html

views/dashboard/dashboard.hbs
```diff
@@ -40,7 +40,7 @@
             <div class="row lesson-cards">
                 <div>
                     {{#each events}}
-                    <a class="lesson-card" {{#if url}}href="{{url}}" {{/if}}{{#if alt}}title="{{alt}}" {{/if}}
+                    <a class="lesson-card" aria-label="{{title}}" {{#if url}}href="{{url}}" {{/if}}{{#if alt}}title="{{alt}}" {{/if}}
```
haniapalac, line 43: Without an href a user won't be able to interact with the link. An empty href is valid html.

haniapalac:
> Bildschirmfoto 2020-07-22 um 15 57 08
> 
> Should the title of the news card be focusable, because at the moment you jump directly to "weiterlesen" without knowing what the news is about... Or maybe the whole card can be the whole card? What do you think?

I agree, it is important to allow the users to know what they are going to read about. Also usually such links like "read more" are not very accessible, it is good to give more information in the aria-label

> Bildschirmfoto 2020-07-22 um 15 57 08
> 
> Should the title of the news card be focusable, because at the moment you jump directly to "weiterlesen" without knowing what the news is about... Or maybe the whole card can be the whole card? What do you think?

I fully agree. According to WCAG these kind of link texts are not very helpful when using Assistive Technology. https://webaim.org/techniques/hypertext/link_text#uninformative You need to add an aria-label here that would explain in detail what this is about and warn a user if it redirects to a new page etc
image

views/dashboard/components/homework_tiles.hbs
pawelchoniawko: When tabbing voiceover says it's on a link and you have to press enter to click it, some voiceovers have an info about redirect and opening a new tab. Why should we add more info in this particular case?

luizzaly:
Bildschirmfoto 2020-07-29 um 13 10 57

Can we somehow mention what the article is about? I think from a user perspective it is not valuable to just hear "weiterlesen" when you don't even know what you will read about...

views/dashboard/components/homework_tiles.hbs
haniapalac: This is due to the fact that not all screen readers might announce that it will open in the new window. 
I think this is a nice example of this warning:
image

views/dashboard/dashboard.hbs
haniapalac: Actually I take back my comment about empty href. This website suggests that we should not use them: https://webaim.org/techniques/keyboard/
image
We can use something like href="#0" etc. instead.

luizzaly (approved):

PR #2076, opened 2020-08-06
Quickfix - missing interpolation

ABialoskorski:
# Description
fixed missing keys

## Screenshots of UI changes
image

PR #2098, opened 2020-08-16
i18n SC-6153 Quickfix invitation modal

DevSchmidtchen:
Ticket: https://ticketsystem.schul-cloud.org/browse/SC-6153

When inviting students there were two wrong keys (headline and below).

**Affected Pages**

- `administration/students/_id/edit`

**Before**
Before

**After**
After


ABialoskorski:
fixed already here https://github.com/hpi-schul-cloud/schulcloud-client/pull/2076

PR #2099, opened 2020-08-17
i18n SC-6161 Replacing "heading" with "headline"

DevSchmidtchen:
Ticket: https://ticketsystem.schul-cloud.org/browse/SC-6161

There were some keys with the wrong naming ("heading" instead of "headline")


ABialoskorski (approved):

DanielRieckhof (approved):

monaweitzenberg (approved):
:)

PR #2071, opened 2020-08-05
SC-5915: a11y homework details

haniapalac:
# Description

The dropdown issues will be fixed by these tickets: SC-5952 and SC-5858
the issue with weird focus behavior for upload section should be handled by this ticket: SC-5957
issues with the ckeditor should also be handled separately.

1. Labels are associated to the input fields (except input fields created with external libraries -> these should be handled in a different ticket)
2. Contrast issues are fixed.
3. file upload is accessible with a keyboard

## Links to Tickets or other pull requests

https://ticketsystem.schul-cloud.org/browse/SC-5915

## Changes

image

## Approval for review
- [x] All points were discussed with the ticket creator, support-team or product owner. The code upholds all quality guidelines from the PR-template.

> Notice: Please remove the WIP label if the PR is ready to review, otherwise nobody will review it.

### Link to Definition of Done
More and detailed information on the *definition of done* can be found [on Confluence](https://docs.schul-cloud.org/pages/viewpage.action?pageId=92831762)

PR #2110, opened 2020-08-18
SC-6120: files a11y

Tratori:
# Description
https://ticketsystem.schul-cloud.org/browse/SC-6120

- folders and files now accessible by keyboard
- some aria-labels are added
- readability of table is enhanced for screen readers
- some labels are linked to inputs
- added a missing translation (i18n)


views/files/file-viewer.hbs
```diff
@@ -1,16 +1,16 @@
 <section class="file-viewer">
-    <a href="#" class="lightbox videostop" id="file-view">
+    <a href="#" class="lightbox videostop" id="file-view" role="button" aria-label="{{$t "files.label.backToFiles"}}">
       <video id="my-video" class="video-js vjs-default-skin videoclick" autoplay controls preload="auto" data-setup='{"fluid": true}'></video>
 		  <script src="/vendor-optimized/video.js/dist/video.min.js" nonce="{{nonceValue}}"></script>
 		  <script src="/vendor-optimized/video.js/dist/lang/{{ @root.userLanguage }}.js" nonce="{{nonceValue}}"></script>
-      <img id="picture">
+      <img id="picture" alt="">
```
luizzaly, line 6: can we add a meaningful alt here or use role="presentation"? :)

views/files/files-grid.hbs
```diff
@@ -26,43 +28,53 @@
 								{{#ifeq this.securityCheck.status "blocked"}}
 									<a href="#"
 									class="btn-file-danger"
-									title="{{$t "files.label.dangerousCode"}}">
+									title="{{$t "files.label.dangerousCode"}}"
```
luizzaly, line 31: Do we need the title here when we have an aria-label as well?

views/files/files-grid.hbs
```diff
@@ -26,43 +28,53 @@
...
+									aria-label="{{$t "files.label.dangerousCode"}}">
 										<i class="fa fa-exclamation-triangle"></i>
 									</a>
 								{{else}}
 									<a href="/files/file?download=true&file={{../this._id}}&name={{../this.storageFileName}}"
 										target="_blank"
 										data-method="download"
-										title="{{$t "files.label.downloadFile"}}">
+										title="{{$t "files.label.downloadFile"}}"
```
luizzaly, line 39: same here?  :)

views/files/files-grid.hbs
```diff
@@ -26,43 +28,53 @@
...
+									class="fa fa-edit file-name-edit"
+									role="button">
+								</a>	
 
 								{{#ifneq this.context 'geteilte Datei'}}
 								{{#userHasPermission "FILE_DELETE"}}
 								<a href="/files/file/"
 								   target="_blank"
 								   data-method="delete"
 								   title="{{$t "files.label.deleteFile"}}"
```
luizzaly, line 61: & here

views/files/files-grid.hbs
```diff
@@ -26,43 +28,53 @@
...
 										<i class="fa fa-cloud-download"></i>
 									</a>
 								{{/ifeq}}
 
-								<i  data-file-id="{{this._id}}"
+								<a  href="#"	
+								    data-file-id="{{this._id}}"
 									data-file-name="{{this.name}}"
 									data-file-path="{{this.path}}"
 									title="{{$t "files.label.renameFile"}}"
```
luizzaly, line 50: & here ..

luizzaly (commented):
I can't upload a file via keyboard, if you want to fix it here maybe have a look how Hanna fixed it in the homework section in the follwing PR: https://github.com/hpi-schul-cloud/schulcloud-client/pull/2071.

In general I was wondering if we need the title when having a aria-label ? :)

views/files/file-viewer.hbs
Tratori: A (more or less) meaningful alt is added when the img is loaded (see files.js changes)
Will that do? :)

views/files/files-grid.hbs
Tratori: The Title is needed for the pop-up afaik:
title
Is the title sufficient? Should I remove the aria-label then?

Tratori:
> I can't upload a file via keyboard, if you want to fix it here maybe have a look how Hanna fixed it in the homework section in the follwing PR: #2071.
> 
> In general I was wondering if we need the title when having a aria-label ? :)



> I can't upload a file via keyboard, if you want to fix it here maybe have a look how Hanna fixed it in the homework section in the follwing PR: #2071.
> 
> In general I was wondering if we need the title when having a aria-label ? :)

It works for me. Perhaps the "old" files.js was still loaded? If not: can you tab the field and does anything else happen on click / enter?

views/files/file-viewer.hbs
```diff
@@ -1,16 +1,16 @@
 <section class="file-viewer">
-    <a href="#" class="lightbox videostop" id="file-view">
+    <a href="#" class="lightbox videostop" id="file-view" role="button" aria-label="{{$t "files.label.backToFiles"}}" tabindex="0">
```
haniapalac, line 2: instead of adding a role and tabindex just use button :-)

views/files/file-viewer.hbs
haniapalac: Why don't we add it directly here? It looks like it is just an i18n key.

views/files/file-viewer.hbs
```diff
@@ -1,16 +1,16 @@
...
-      <img id="picture">
+      <img id="picture" alt="">
     </a>
 </section>
 
 {{#embed "lib/components/modal" class="open-modal" title=($t "files.headline.doYouWantToOpenThisFile")}}
     {{#content "body"}}
         <div class="modal-body">
-            <input class="box" type="checkbox" name="allowExternalViewers" value="save">{{$t "files.text.rememberThisSelection"}}<br>
+            <input class="box" type="checkbox" name="allowExternalViewers" value="save" aria-label="{{$t "files.text.rememberThisSelection"}}">{{$t "files.text.rememberThisSelection"}}<br>
```
haniapalac, line 13: actually we don't really need an aria-label here because we have a visual label which is exactly the same. It would make sense to override the visual label for screen readers if additional information was added/needed for the screen reader users.

views/files/files-grid.hbs
```diff
@@ -8,7 +8,9 @@
 							{{#ifneq this.securityCheck.status "blocked"}} fileviewer{{/ifneq}}"
 							{{#ifneq this.securityCheck.status "blocked"}} data-file-viewer-type="{{../this.type}}"{{/ifneq}}
 							{{#ifneq this.securityCheck.status "blocked"}} data-file-viewer-savename="{{../this.saveName}}"{{/ifneq}}
-							{{#ifneq this.securityCheck.status "blocked"}} data-file-viewer-id="{{../this._id}}"{{/ifneq}}>
+							{{#ifneq this.securityCheck.status "blocked"}} data-file-viewer-id="{{../this._id}}"{{/ifneq}}
```
haniapalac, line 11: it is always better to use a button, if possible, instead of adding a role and tabindex, it's cleaner. Is there a reason why you didn't use a button here?

views/files/files-grid.hbs
haniapalac: Yes, then I guess we don't need an aria-label. we don't need an href for the button. You can also hide the icon from screen readers with aria-hidden.

views/files/files-grid.hbs
haniapalac: It is always good to add a warning for screen reader users when they are redirected to a new page/tab, otherwise they might expect to be still on the same page. So here I would add the warning to the aria-label (the title doesn't need that warning obviously).

views/files/files-grid.hbs
```diff
@@ -24,47 +26,55 @@
...
-										title="{{$t "files.label.downloadFile"}}">
+										title="{{$t "files.label.downloadFile"}}"
+										aria-label="{{$t "files.label.downloadFile"}}">
 										<i class="fa fa-cloud-download"></i>
-									</a>
+									</button>
 								{{/ifeq}}
 
-								<i  data-file-id="{{this._id}}"
+								<button  href="#"
```
haniapalac, line 46: remove the href

views/files/files-grid.hbs
```diff
@@ -24,47 +26,55 @@
...
-									class="fa fa-edit file-name-edit">
-								</i>
+									aria-label="{{$t "files.label.renameFile"}}"
+									class="file-btn fa fa-edit file-name-edit">
+								</button>	
 
 								{{#ifneq this.context 'geteilte Datei'}}
 								{{#userHasPermission "FILE_DELETE"}}
-								<a href="/files/file/"
+								<button href="/files/file/"
```
haniapalac, line 57: we need an anchor here, because we have a meaningful href + see the comment above aobut the warning when opening in the new tab.

views/files/files-grid.hbs
```diff
@@ -24,47 +26,55 @@
...
-								   data-file-name="{{../this.name}}"
+								   data-file-name="{{../this.name}}">
 									<i class="fa fa-trash-o"></i>
-								</a>
+								</button>
 								{{/userHasPermission}}
 
-								<a href="#"
-								class="btn-file-share"
+								<button href="#"
```
haniapalac, line 69: we don't need an href + aria-label because it is the same as the title, you can also hide the icon

views/files/files-grid.hbs
```diff
@@ -76,27 +86,29 @@
 								</div>
 
 									{{#if ../this.userIsOwner}}
-										<a
+										<button
 										href=""
```
haniapalac, line 90: remove the href

views/files/files-grid.hbs
```diff
@@ -76,27 +86,29 @@
...
-										class="btn-file-move"
+										class="file-btn btn-file-move"
 										title="{{$t "files.label.moveFile"}}"
+										aria-label="{{$t "files.label.moveFile"}}"
```
haniapalac, line 93: we can skip the aria-label here

views/files/files-grid.hbs
```diff
@@ -76,27 +86,29 @@
...
+										data-disabled="true">
 											<i class="fa fa-folder-open"></i>
-										</a>
+										</button>
 									{{/if}}
 								{{/ifneq}}
 
-								<a
+								<button
 									href="#"
```
haniapalac, line 104: remove the href

views/files/files-grid.hbs
```diff
@@ -76,27 +86,29 @@
...
-									class="btn-file-settings"
+									class="file-btn btn-file-settings"
 									title="{{$t "files.label.editFilePermissions"}}"
+									aria-label="{{$t "files.label.editFilePermissions"}}"
```
haniapalac, line 107: we can skip the aria-label here

views/files/files.hbs
```diff
@@ -57,8 +57,7 @@
 				<div class="row">
 					{{#each directories}}
 						<div class="col-xs-12">
-						  <div class="card card-block folder openfolder" data-folder-id="{{this._id}}">
-                              <button type="button" class="button-transparent">
+						  <div class="card card-block folder openfolder" data-folder-id="{{this._id}}" tabindex="0" role="button" aria-label="{{$t "files.label.clickToOpenDir" (dict "name" this.name)}}">
```
haniapalac, line 60: better to make it a button and restyle it with css

haniapalac (changes requested):

haniapalac:
One more thing: it would be good to include the name of the card in the aria-label of the button so the user knows what type of files they will open:
image

views/files/file-viewer.hbs
Tratori: This would destroy the ability to display videos, as when loading a video, buttons will be added as children nodes to this node which wouldnt be valid html.

views/files/file-viewer.hbs
Tratori: this alt might be a little unnecessary anyways: When clicking on an image in your files it is displayed here. This alt is loaded at that moment and reads something like this: "The Image picture.png"

views/files/file-viewer.hbs
Tratori: done

views/files/files-grid.hbs
Tratori: This node contains other buttons, so it can't be changed to a button unfortunately.

views/files/files-grid.hbs
Tratori: done

locales/de.json
```diff
@@ -1316,14 +1319,20 @@
 			"doYouWantToOpenThisFile": "Möchtest du diese Datei mit einem externen Dienst öffnen?",
 			"newFolder": "Neuer Ordner"
 		},
+		"img_alt":{
+			"theImageX" : "Das Bild {{imgName}}"
```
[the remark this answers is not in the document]
Tratori: done

views/files/files-grid.hbs
Tratori: The href is only the url needed for the fetch to the server, so there is no "real link" here as no new tab is opened

views/files/files.hbs
Tratori: done

views/files/files-grid.hbs
haniapalac: that sounds like bad html. Do we need to have the path as an href attribute on the element? Can't we save the path somewhere else?

haniapalac (approved):

PR #2116, opened 2020-08-20
SC-5710 courses a11y

pawelchoniawko:
# Description
A11y courses

## Links to Tickets or other pull requests
https://ticketsystem.schul-cloud.org/browse/SC-5710

PR #2126, opened 2020-08-21
A11y Refinement

Tratori:
# Description
https://ticketsystem.schul-cloud.org/browse/SC-6339

Fixing Typos and broken Keys regarding a11y & i18n


qnowak:
We do not need this img_alt.brokenPen because it brings nothing to the user using the reader. In this https://github.com/hpi-schul-cloud/schulcloud-client/pull/2116 PR all img_alt.brokenPen occurrences are deleted.

luizzaly (approved):

PR #2153, opened 2020-09-01
i18n SC-6529 Wrong language in team invitation

DevSchmidtchen:
Ticket: https://ticketsystem.schul-cloud.org/browse/SC-6529

**Affected Pages**
`/teams > open team detail view > click three dots next to team name > "Teammitglieder verwalten" > "Externe Teilnehmer einladen"`

**Before**
Before Screenshot

**After**
After Screenshot
After Screenshot Mobile View

### DIFF
--- a/locales/de.json
+++ b/locales/de.json
@@ -1808,8 +1808,8 @@
 		},
 		"help_fromInviteExternalMember": {
 			"button": {
-				"externalExpert": "External experts",
-				"teachersFromOtherSchool": "Teachers from other schools"
+				"externalExperts": "Externe Experten",
+				"teachersFromOtherSchool": "Lehrer von anderen Schulen"
 			},
 			"heading": {
 				"whoDoYouWantToInviteTheTeam": "Wen möchtest du ins Team einladen?"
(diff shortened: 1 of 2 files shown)


janrenz (approved):

Parnswir (approved):

monaweitzenberg (changes requested):
image
please change Button Text into: "Via E-Mail einladen" 
thx :)

monaweitzenberg (changes requested):
Space between buttons would be helpful for the mobile-user, what do you think?

DevSchmidtchen:
> Space between buttons would be helpful for the mobile-user, what do you think?

Generally a great idea but this would not only affect this screen. The spacing of the buttons is a global thing and therefore it would change the spacing of every button in mobile view. I think I shouldn't do this in this specific PR.

ABialoskorski (approved):

ABialoskorski:
> Space between buttons would be helpful for the mobile-user, what do you think?

this is not part of this ticket, please create a ticket for that

monaweitzenberg (approved):

PR #2168, opened 2020-09-06
i18n SC-6641 improvements

DevSchmidtchen:
Ticket: https://ticketsystem.hpi-schul-cloud.org/browse/SC-6641


ABialoskorski:
as the requirements changed, please add english translations also :)

ABialoskorski:
we need to check how big scope this should have, beacuse we can have this PR forever :D

DevSchmidtchen:
> we need to check how big scope this should have, beacuse we can have this PR forever :D

You are probably right, I am currently revisiting every key in `de.json` and this should be finished very soon and then I would change label from `WIP` to `waiting for review`.

ABialoskorski (approved):

monaweitzenberg (approved):

DanielRieckhof (approved):

PR #2183, opened 2020-09-10
SC-6714 - max size in MB fix

ABialoskorski:
# Description
fixed max size of file in MB
user can't upload a file because max size was not visible for .hbs so it was 0MB.

- https://ticketsystem.schul-cloud.org/browse/SC-6714

## Screenshots of UI changes
no UI changed


DanielRieckhof (approved):
And this was the only reason why the upload failed? 
... Wow! :-)

monaweitzenberg (approved):

PR #2237, opened 2020-09-29
Better handling of shared files and UX improvements on files

janrenz:
Force download for shared files.


janrenz:
Bildschirmfoto 2020-09-29 um 22 17 57
I decided to change the fix and instead just add the file to the shared files are and not open it without additional user interaction.
So this is how it works now: User clicks or open share link. User is redirect to his shared files folder, message appears

janrenz:
Along the way I found a bug and fixed that so the share button is not shown for shared files

janrenz:
I also disabled the edit permissions button in areas where the click on the edit permissions button did not worked.

Parnswir (approved):

MartinSchuhmacher (approved):

PR #2249, opened 2020-10-07
SC-7082 - csrf token required when loging out

ABialoskorski:
# Description
addec check if csrf token exists when user try to logout

- https://ticketsystem.schul-cloud.org/browse/SC-7082


mwolowyk-cap (changes requested):
Looks good for me. Please fix codacy warning.

controllers/login.js
```diff
@@ -213,12 +215,21 @@ router.get('/login/systems/:schoolId', (req, res, next) => {
...
-	api(req).del('/authentication') // async, ignore result
-		.catch((err) => { logger.error('error during logout.', { error: err.toString() }); });
-	return authHelper
-		.clearCookie(req, res, { destroySession: true })
-		.then(() => res.redirect('/'))
-		.catch(next);
+	if (res.locals.csrfToken) {
+		api(req).del('/authentication') // async, ignore result
+			.catch((err) => { logger.error('error during logout.', { error: err.toString() }); });
+		authHelper
```
mwolowyk-cap, line 221: I don't think that the function will exit at that point, In this case the function will return null and not the result of the authHelper. If the result of the function is not used why to return null?

mwolowyk-cap (changes requested):

controllers/login.js
ABialoskorski: yeah good point, my mistake

ABialoskorski: done ;)

mwolowyk-cap (approved):

controllers/login.js
```diff
@@ -213,12 +215,19 @@ router.get('/login/systems/:schoolId', (req, res, next) => {
 });
 
 router.get('/logout/', (req, res, next) => {
-	api(req).del('/authentication') // async, ignore result
-		.catch((err) => { logger.error('error during logout.', { error: err.toString() }); });
-	return authHelper
-		.clearCookie(req, res, { destroySession: true })
-		.then(() => res.redirect('/'))
-		.catch(next);
+	if (res.locals.csrfToken) {
```
DanielRieckhof, line 218: But I think, this will be always fulfilled. We need to add the csrf token to the request and check it, like here:
image

DanielRieckhof: But maybe I didn't get, how this csrf token is created and handled ...

PR #2263, opened 2020-10-15
SC-5390 Fixed static texts

janrenz:
Use existing Config var to render texts


controllers/administration.js
```diff
@@ -1211,6 +1211,7 @@ router.get(
 					user,
 					password: authHelper.generateConsentPassword(),
 					referrer: req.header('Referer'),
+					age: CONSENT_WITHOUT_PARENTS_MIN_AGE_YEARS,
```
CordlessWool, line 1214: age is littlebit to generic, please a more self speaking name for the attribute

CordlessWool (changes requested):

views/administration/users_skipregistration.hbs
```diff
@@ -57,7 +57,7 @@
 
                     {{#unless editTeacher}}
                     <div class="col-md-6">
-                        {{$t "administration.global.text.forStudentsUnder" (dict "minAge" CONSENT_WITHOUT_PARENTS_MIN_AGE_YEARS )}}<br />
+                        {{$t "administration.global.text.forStudentsUnder" (dict "minAge" age )}}<br />
```
CordlessWool, line 60: please use same placeholder on the text, so it is more clear if someone taking a look on it

CordlessWool (changes requested):

controllers/administration.js
janrenz: age is already used in other places for the same var, so while it may make sense to unify this but is out of scope for this tickt.

views/administration/users_skipregistration.hbs
janrenz: age is already used in other places for the same var, so while it may make sense to unify this but is out of scope for this tickt.

controllers/administration.js
tofixx, line 1214:
```suggestion
					minAge: CONSENT_WITHOUT_PARENTS_MIN_AGE_YEARS,
```

views/administration/users_skipregistration.hbs
tofixx, line 60:
```suggestion
                        {{$t "administration.global.text.forStudentsUnder" (dict "age" minAge )}}<br />
```

views/administration/users_skipregistration.hbs
```diff
@@ -73,7 +73,7 @@
 						</label><br />
                     </div>
                     <div class="col-md-6">
-                        {{$t "administration.users_skipRegistration.text.forStudentsAged14" }}<br />
+                        {{$t "administration.users_skipRegistration.text.forStudentsAgedLimit" (dict "age" age)}}<br />
```
tofixx, line 76:
```suggestion
                        {{$t "administration.users_skipRegistration.text.forStudentsAgedLimit" (dict "age" minAge)}}<br />
```

tofixx (commented):
use consistent naming. in controller i would use minAge, for replacements in translations I would allow age, or use the same name for everything

PR #2349, opened 2020-11-20
Hotfix/SC-1589 - Update error handling on team creation

DanielRieckhof:
# Description
We have already updated the handling of name and description string within team model. But the error handling on legacy client only shows an internal error when creating a team with empty name. The error handling on team update was correct. 

https://ticketsystem.hpi-schul-cloud.org/browse/SC-1589

## Changes
Use the same error handling on team creation as on team editing

## Approval for review

### DIFF
--- a/controllers/teams.js
+++ b/controllers/teams.js
@@ -338,41 +338,12 @@ router.get('/', async (req, res, next) => {
 	}
 });
 
-router.post('/', async (req, res, next) => {
-	// map course times to fit model
-	(req.body.times || []).forEach((time) => {
-		time.startTime = moment.duration(time.startTime, 'HH:mm').asMilliseconds();
-		time.duration = time.duration * 60 * 1000;
-	});
-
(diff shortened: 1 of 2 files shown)


mwolowyk-cap (approved):

controllers/teams.js
```diff
@@ -365,17 +365,14 @@ router.post('/', async (req, res, next) => {
 		delete req.body[feature];
 	});
 
-	api(req)
```
tofixx, line 368: an await in front of the api call and the removal of the catch L375-377 should have the same efect than adding the try/catch. however the next middleware will be called when an error is thrown. because of this the try/catch should not be required and i would just add an await and remove the catch which swallows the error.
If there is an error thrown above within of this method, the error will be catched in the next middleware too.

tofixx (changes requested):

tofixx (approved):

PR #2350, opened 2020-11-20
async middleware error handler added (WIP)

tofixx:
# Description

This PR is fixing the control flow for error handling within of async middleware functions.

## Approval for review


virgilchiriac (approved):

controllers/alerts.js
```diff
@@ -1,22 +1,20 @@
...
-	handler = (req, res) => api(req).get('/alert')
-		.then((alert) => {
-			res.json(alert);
-		}).catch((err) => {
-			logger.error(new Error('Can not get /alert', err));
-			res.json([]);
-		});
-}
-router.get('/', handler);
+router.get('/', asyncWrapper(async (req, res, next) => {
```
CeEv, line 9: I prefer the old solution without wrapper and added catch with different handler for set feature flag and without. At the moment the wrapper is not needed at this place.

middleware/asyncMiddlewareWrapper.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Express is loosing the control flow for error handling of async middleware functions.
+ * All async middlewares need to be wrapped with this method to ensure error handling works as expected.
+ * @param {*} middlewareFunction async (req, res, next) - next must be the last argument
+ */
+module.exports = (middlewareFunction) => function asyncWrapper(...args) {
```
CeEv, line 6: At the moment we have no place where it is needed, but any async handler include 3 arguments req, res, next. I see no reason for make the code more complicated than it needs to be.
The second point that we should discuss is, if it is the solution. 
If i know that i must use this for catching errors, that i know that i must add the catch and add this. 
If i do not know it, i do not use this wrapper. 
It not prevent us for this error mistakes.

CeEv (changes requested):

PR #2370, opened 2020-11-30
SC-7983 - fixed that topic is now choosable after select of course

ABialoskorski:
# Description
topic was empty when creating homework, even when there were topics

- https://ticketsystem.schul-cloud.org/browse/SC-7983


views/homework/edit.hbs
```diff
@@ -54,6 +54,7 @@
 							$('#lessonPicker').append('<option value="' + r.lessons.data[i]._id + '">' + r.lessons.data[i].name + '</option>');
 						}
 						$('#lessonPicker').prop('disabled', false);
+						$('#lessonPicker').trigger("chosen:updated");
```
mwolowyk-cap, line 57: you should update the picker also for the case when there were no tasks found.

mwolowyk-cap: I would put it to line 60 and also probably to line 64. Please check whether it is required.

mwolowyk-cap (changes requested):

views/homework/edit.hbs
ABialoskorski: done ;)

mwolowyk-cap (approved):

PR #2390, opened 2020-12-08
Feature/sc 6563 moving sections in themes only moves heading

robertstrz:
Even though  changes has been mode only in one of the n editors, the following callback is executed for some reason for all opened editors. That is why it overwrites images in another editors as well. 
```
editor.model.change((writer) => {
```
Proper fix would be to execute aforementioned callback only for the editor that is currently "touched". Fix introduced in this PR is the only solution, that I came up with and couldn't implement suggested solution.

# Description

https://ticketsystem.hpi-schul-cloud.org/browse/SC-6563

## Approval for review


static/scripts/topicEdit.js
```diff
@@ -7,6 +7,7 @@ import {
 } from 'react-sortable-hoc';
 import './calendar';
 import ClassicEditor from '@ckeditor/ckeditor5-editor-classic/src/classiceditor';
+import shortid from 'shortid';
```
alriese, line 10: I think `shortid` is deprecated and should not be used anymore (https://www.npmjs.com/package/shortid). The authors of shortid recommend 'nanoid' instead. (https://github.com/ai/nanoid/)

static/scripts/topicEdit.js
```diff
@@ -184,7 +185,7 @@ const SortableList = SortableContainer(({
         <div>
             {items.map((value, index) => (
                 <SortableItem
-                    key={`item-${index}`}
+                    key={`item-${value._id ? value._id : value.key}`}
```
alriese, line 188: Why do we need the distinction? In which cases do we have an `_id` but not an `key` or or vice versa?

static/scripts/ckeditor/plugins/ckeditor5-filebrowser/src/filebrowser.js
```diff
@@ -87,8 +89,10 @@ export default class FileBrowserPlugin extends Plugin {
 							src: imageUrl,
 							alt: imageAltText,
 						});
-
-						editor.model.insertContent(imageElement, editor.model.document.selection);
+						const lastOpenedEditorId = document.getElementById('editor-id').value;
```
alriese, line 92: It seems that there must be a nicer way to find out the current editor, but I couldn't find one now either. If it works I would go with it for now.

static/scripts/topicEdit.js
robertstrz: If you add new editor that is not yet saved then there is not _id  but I am assigning id with the help of shortid library. Then when you saved it and reload e.g. the value._id is present since it comes from the backend

static/scripts/topicEdit.js
robertstrz: Yeah, it was deprecated due to the cryptographics reasons. In this case it is just used to generate random id for newly added  editor. I don't care if someone will guess this Id because there is no harm behind it. Currently it is used in some places in our code. I would recommend to create a follow up ticket (I can do it) to replace shortid with nano id what do you say ?

static/scripts/ckeditor/plugins/ckeditor5-filebrowser/src/filebrowser.js
robertstrz: Me either...

static/scripts/topicEdit.js
alriese: Sounds good 👍

static/scripts/topicEdit.js
alriese: Thank you for the clarification

alriese (approved):

PR #2411, opened 2020-12-16
SC-8246 Asset and route caching 

janrenz:
This will cache all assets that are using the helper in the browser Cache. If the file is changed the helper will create a new URL in the template.
It works like this: 

Files where the new helper Method is used to load are fingerprinted and will have a very long local cache time.
http://localhost:3100/scripts/messenger/welcome.**503c24f**.js
If the file content changes the hash param also changes.
The hashes are calculated on app load and stored oil a in memory mapping table.

Other files from the static dir will have a one day local cache. This is a fallback method. If this files are changed they will live for up to one day in the cache of the browser, so there this must be ok or they must be using a version param in the path.

Previously we used the standard ETAG way of caching, which still produces HTTP round trips and load on the clients.

For local development this caching behavior can be turned using  FEATURE_ASSET_CACHING_ENABLED

This PR also introduces a 90sec ones client cache for the alerts route.

To not disrupt developer workflows the Feature flipper is now off in dev by default.


maxklenk:
Nice approach, to keep the generated assets as they are, but load them using different names.`staticify` caches the generated hashes so its performance should be good.

The only problem I see is, that all asset requests have to go through the node stack to rewrite the url. I would expect that assets are served directly by nginx. If the hashes were added as query string instead caching would be still possible and the files could be served directly by a proper webserver.

A simple and not as efficient caching mechanism could be:
```node
const package = require('../../../package.json');
getVersionedPath: (value) => {
    if (value.includes('?')) {
	return value + '&v=' + package.version;
    } else {
	return value + '?v=' + package.version;
    }
}
```

theme/open/views/lib/default.hbs
```diff
@@ -41,13 +41,13 @@
...
 
 	{{#block "styles"}}
-	<link rel="preload" href="/fonts/ptsans/ptsans-regular.woff2" as="font" type="font/woff2" crossorigin>
-	<link rel="preload" href="/fonts/ptsans/ptsans-bold.woff2" as="font" type="font/woff2" crossorigin>
-	<link rel="preload" href="/fonts/ptsans/ptsans-italic.woff2" as="font" type="font/woff2" crossorigin>
-	<link rel="preload" href="/vendor-optimized/font-awesome/fonts/fontawesome-webfont.woff2?v=4.7.0" as="font"
+	<link rel="preload" href="{{getVersionedPath '/fonts/ptsans/ptsans-regular.woff2'}}" as="font" type="font/woff2" crossorigin>
+	<link rel="preload" href="{{getVersionedPath '/fonts/ptsans/ptsans-bold.woff2'}}" as="font" type="font/woff2" crossorigin>
+	<link rel="preload" href="{{getVersionedPath '/fonts/ptsans/ptsans-italic.woff2'}}" as="font" type="font/woff2" crossorigin>
+	<link rel="preload" href="{{getVersionedPath '/vendor-optimized/font-awesome/fonts/fontawesome-webfont.woff2?v=4.7.0'}}" as="font"
```
maxklenk, line 47: The used fonts are defined in `ptsans.css`. Since staticify does not update the content of these files, I would suggest to manually add a version query(`?v=1.0.0` ) to the font files and inside the css.

views/about/about_loggedin.hbs
```diff
@@ -1,13 +1,13 @@
 {{#extend "lib/loggedin" hideSearch="true"}}
 
 {{#content "styles" mode="append"}}
-<link rel="stylesheet" href="/styles/partner/partner.css" />
-<link rel="stylesheet" href="/styles/about/about.css" />
+<link rel="stylesheet" href="{{getVersionedPath '/styles/partner/partner.css'}}" />
+<link rel="stylesheet" href="{{getVersionedPath /styles/about/about.css'}}" />
```
maxklenk, line 5: missing `'`

janrenz: fixed

janrenz:
Maybe we should add a feature flipper for this., so we can turn it off if it makes issues

theme/open/views/lib/default.hbs
janrenz: I added a default cache for other assets

helpers/handlebars/helpers/index.js
```diff
@@ -177,6 +187,12 @@ const helpers = () => ({
 		return options.inverse(this);
 	},
 	userIds: (users) => (users || []).map((user) => user._id).join(','),
+	getVersionedPath: (assetPath) => {
```
tofixx, line 190: this is just getAssetPath, if it's versioned or not just will be changed inside.
```suggestion
	getAssetPath: (assetPath) => {
```

helpers/handlebars/helpers/index.js
```diff
@@ -6,6 +6,16 @@ const { Configuration } = require('@hpi-schul-cloud/commons');
...
 const Globals = require('../../../config/global');
+const path = require('path');
+const fs = require('fs');
+
+let themeAssetDir = path.join(__dirname, '../../..', `build/${Globals.SC_THEME}`);
+if (!fs.existsSync(themeAssetDir)) {
+	// fallback to static dir, for example for test env
+	themeAssetDir = path.join(__dirname, '../../..', 'static');
+}
+const staticify = require('staticify')(themeAssetDir);
```
tofixx, line 17: creation of static files should happen in build process is this possible? 
all running pods need to have the same hash (ok, it's md5)...

app.js
```diff
@@ -127,9 +127,20 @@ if (Configuration.get('FEATURE_MORGAN_LOG_ENABLED')) {
 app.use(bodyParser.json({ limit: '10mb' }));
 app.use(bodyParser.urlencoded({ extended: true, limit: '10mb' }));
 app.use(cookieParser());
-app.use(express.static(path.join(__dirname, `build/${themeName}`)));
+
 app.use('/locales', express.static(path.join(__dirname, 'locales')));
```
tofixx, line 131: no caching for locales?

app.js
```diff
@@ -127,9 +127,20 @@ if (Configuration.get('FEATURE_MORGAN_LOG_ENABLED')) {
...
 
+let themeAssetDir = path.join(__dirname, `build/${themeName}`);
```
tofixx, line 133: remove code duplication with https://github.com/hpi-schul-cloud/schulcloud-client/pull/2411/files#diff-86ec246c0db98dcda6ecd01fd13b114a98f6455fdd661e16324569e0064c0afbR12

app.js
```diff
@@ -127,9 +127,20 @@ if (Configuration.get('FEATURE_MORGAN_LOG_ENABLED')) {
...
+if (!fs.existsSync(themeAssetDir)) {
+	// fallback to static dir, for example for test env
+	themeAssetDir = path.join(__dirname, 'static');
+}
+const staticify = require('staticify')(themeAssetDir, { maxAgeNonHashed: '1d' });
```
tofixx, line 138: do not create a separate instance of staticify here and in the handlebars helper again, use it once

app.js
```diff
@@ -127,9 +127,20 @@ if (Configuration.get('FEATURE_MORGAN_LOG_ENABLED')) {
...
+app.use(staticify.middleware);
+
+// backup
+app.use(express.static(path.join(__dirname, `build/${themeName}`)));
```
tofixx, line 142: keep this line where it was before

helpers/handlebars/helpers/index.js
```diff
@@ -6,6 +6,16 @@ const { Configuration } = require('@hpi-schul-cloud/commons');
 const permissionsHelper = require('../../permissions');
 const i18n = require('../../i18n');
 const Globals = require('../../../config/global');
+const path = require('path');
+const fs = require('fs');
+
+let themeAssetDir = path.join(__dirname, '../../..', `build/${Globals.SC_THEME}`);
```
tofixx, line 12: do not use globals
```suggestion
let themeAssetDir = path.join(__dirname, '../../..', `build/${Configuration.get('SC_THEME')}`);
```

app.js
tofixx, line 138: in other place a feature flag has been used which should be used here too.

tofixx (changes requested):

helpers/handlebars/helpers/index.js
janrenz: it versioned in a sense that it contains the fingerprint. so no server, but rather a hash..

helpers/handlebars/helpers/index.js
janrenz: this just uses the files already generated in the build process., so they are the same on all pods. 

There might be edgewise during rolling rollouts, but this case be investigated later

app.js
janrenz: code optimization can be handled in a follow up, I will be busy with devops and org stuff till jan.

app.js
janrenz: locales are also cached, worked out of the box.

app.js
janrenz: feel free to optimize in this pr.

app.js
janrenz: actually this is not really need and it has to come after the staticify habnlder, otherwise the fallback cache for assets is not working

helpers/handlebars/helpers/index.js
janrenz: can you change this?

app.js
janrenz: same here.

helpers/handlebars/helpers/index.js
janrenz: but I would be ok with renaming

tofixx (approved):

tofixx (approved):

theme/open/views/lib/default.hbs
maxklenk: The cache headers are not added for the fonts at the moment, but preloading the font using a different name, than the name defined in css results in loading the fonts two times. I fixed that.

maxklenk (approved):
I fixed some problems and think this should work.

After the first deployment and tests we should increase the caching duration:
Suggested Duration: 1 Year
https://web.dev/uses-long-cache-ttl/?utm_source=lighthouse&utm_medium=devtools#how-to-cache-static-resources-using-http-caching

middleware/assets.js
```diff
@@ -0,0 +1,56 @@
...
+	}
+};
+
+expectAssetsDir();
+
+// configure static file hashing and caching
+const staticifyInstance = staticify(buildThemeAssetDir, {
+	maxAgeNonHashed: '1d',
+	sendOptions: {
+		maxAge: 3600 * 1000, // one hour, in milliseconds
```
tofixx, line 27: increase default, move to config

tofixx:
requires
https://github.com/hpi-schul-cloud/end-to-end-tests/pull/274 
https://github.com/hpi-schul-cloud/docker-compose/pull/34 
to be merged

theme/brb/views/authentication/home.hbs
```diff
@@ -1,6 +1,6 @@
 {{#extend "lib/default"}} {{#content "styles" mode="append"}}
-<link rel="stylesheet" href="/styles/authentication/home.css" /> {{/content}} {{#content "scripts" mode="append"}}
-<script src="/scripts/login.js" type="text/javascript" nonce="{{nonceValue}}" defer></script>
+<link rel="stylesheet" href="{{getAssetPath '/styles/authentication/home.css'}}" /> {{/content}} {{#content "scripts" mode="append"}}
+<script src="{{getAssetPath '/scripts/login.js‘}}" type="text/javascript" nonce="{{nonceValue}}" defer></script>
```
alriese, line 3:
```suggestion
<script src="{{getAssetPath '/scripts/login.js'}}" type="text/javascript" nonce="{{nonceValue}}" defer></script>
```

alriese (changes requested):

alriese (approved):

tofixx:
Already merged.

PR #2436, opened 2021-01-16
SC-8416 Fix logo display on subpages by making path absolute

janrenz:
Fixed display of logo on subpages on tablet.

Bildschirmfoto 2021-01-16 um 12 03 56


tofixx (commented):
Are there other absolute pathes that can be found by search that should be updated?

Metauriel (approved):

PR #2442, opened 2021-01-20
Hotfix/1/sc 8189 course event duplication

mwolowyk-cap:
# Description

We have an issue by updating the course with events when we use multiple mongo db replicas due to eventual inconsistency.

When the course is updated (patched) following logic happens: 

1. Delete all events for this course in calendar.
2. Update course
3. Create new events for this course in calendar.

**Solution:** 
Introduce additional step

2.1,  Get updated course from the database again instead of using update result.

https://ticketsystem.hpi-schul-cloud.org/browse/SC-8189

## Approval for review

### DIFF
--- a/controllers/courses.js
+++ b/controllers/courses.js
@@ -771,59 +771,61 @@ router.get('/:courseId/', async (req, res, next) => {
 });
 
 router.patch('/:courseId', async (req, res, next) => {
-	const redirectUrl = req.query.redirectUrl || `/courses/${req.params.courseId}`;
-
-	// map course times to fit model
-	req.body.times = req.body.times || [];
-	req.body.times.forEach((time) => {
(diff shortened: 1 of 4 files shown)


CordlessWool (dismissed):

controllers/courses.js
```diff
@@ -770,7 +770,7 @@ router.get('/:courseId/', async (req, res, next) => {
 	}
 });
 
-router.patch('/:courseId', (req, res, next) => {
+router.patch('/:courseId', async (req, res, next) => {
```
tofixx, line 773: when using async here, we must try/catch the whole body for catch cou may use `catch(next)`, please return res.foo

tofixx (changes requested):

PR #2519, opened 2021-04-21
Hotfix/2/SC-8971 - remove $limit:false

virgilchiriac:
# Description

- https://github.com/hpi-schul-cloud/schulcloud-server/pull/2473
- https://ticketsystem.schul-cloud.org/browse/SC-8971 

## Approval for review

### DIFF
--- a/controllers/schools.js
+++ b/controllers/schools.js
@@ -2,6 +2,7 @@ const express = require('express');
 
 const router = express.Router();
 const api = require('../api');
+const { getApiData } = require('../helpers/apiData');
 
 // schools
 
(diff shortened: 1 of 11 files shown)


helpers/apiData.js
```diff
@@ -0,0 +1,30 @@
+const api = require('../api');
+
+/* @deprecated */
+const getAllPaginatedData = async (req, service, qs) => {
```
tofixx, line 4: i would not prefer a deprecated function but instead why not let us export getApiData as getAllPagesFromApi or somewhat and remove the $limit from the request qs? Then everybody would know what happens...

controllers/schools.js
```diff
@@ -17,9 +18,15 @@ router.get('/', async (req, res, next) => {
 		params.qs._id = { $ne: res.locals.currentSchool };
 	}
 	try {
-		const schools = await api(req).get('/schools/', params);
+		let schools = [];
+		if (req.query.$limit === 'false') {
```
virgilchiriac, line 22: this is needed because in static script teamMembers we have a call to get all data directly to the /schools endpoint.

PR #2520, opened 2021-04-22
move secrets out of default config

tofixx:
# Description
the cookie_secret which is required will be removed from the default config which is loaded in production and must be manually given from now on. 

## Approval for review


package.json
```diff
@@ -3,7 +3,7 @@
 	"version": "26.1.0",
 	"private": true,
 	"scripts": {
-		"start": "node ./bin/www",
```
tofixx, line 6: this will produce `      message: "should have required property 'COOKIE_SECRET'"` when start without the required setting, we must ensure that all settings from development are defined in production...

PR #2521, opened 2021-04-26
Update wordlist.js

konstantin-sc:
removed the word "gestorben" out of the wordlist because it caused some untasty combinations. please review pragmatically


virgilchiriac (approved):

tofixx (approved):
Good job, we should remove the whole list and change over to random passwords. Next the Innenminister will Flucht ...

mwolowyk-cap (approved):

PR #2656, opened 2021-08-20
bc-8 update dependencies

MartinSchuhmacher:
# Description

the dependency update with commit 48d808273d83121c4bc10768dff5eff568a1adaa also included a winston update, that caused the issue.
that is though winston and request-promise share the same transitive dependency: async.
since request-promise is deprecated we should remove it, see bc-93.

https://ticketsystem.hpi-schul-cloud.org/browse/BC-8

## Changes

- dependency updates
- set language for the user while saving rest of the account data

## Approval for review


.github/workflows/tag.yml
```diff
@@ -14,23 +14,14 @@ jobs:
         tennens: [default, brb, demo, int, n21, open, thr ]
     steps:
       - uses: actions/checkout@v2
```
tofixx, line 16: why changing the workflows, how the relate to dependency updates?

package.json
```diff
@@ -61,102 +61,102 @@
...
 		"feedr": "^2.13.5",
 		"font-awesome": "^4.7.0",
 		"form-serializer": "^2.5.0",
-		"generate-password": "^1.5.1",
+		"generate-password": "^1.6.1",
 		"git-repo-info": "^2.1.0",
 		"gulp-change": "^1.0.2",
 		"gulp-cli": "^2.3.0",
 		"gulp-rename": "^2.0.0",
-		"handlebars": "^4.7.6",
```
tofixx, line 91: what is the dependency which crashed the client? that one should be changed only and it's version should be fixed (to noch change by itself)

tofixx (changes requested):

package.json
MartinSchuhmacher: look into the ticket in the description https://ticketsystem.hpi-schul-cloud.org/browse/BC-8

.github/workflows/tag.yml
Loki-Afro: ignore it for now, this is not finished yet.

tofixx: when this is not finished --> it should not be merged. 
when this is not part of the ticket --> it should be a separate pr.

Metauriel (approved):

PR #2657, opened 2021-08-23
BC-107 improve Dockerfile and pipeline

Loki-Afro:
improve Dockerfile and pipeline


MartinSchuhmacher (approved):
quite ok I would say ;)

PR #2660, opened 2021-08-25
bc-8 update dependencies

Loki-Afro:
the dependency update with commit 48d808273d83121c4bc10768dff5eff568a1adaa also included a winston update, that caused the issue.
that is though winston and request-promise share the same transitive dependency: async.
since request-promise is deprecated we should remove it, see bc-93.


MartinSchuhmacher (approved):
very good implementation

PR #2822, opened 2022-04-08
BC-1393 - Replace current course detail page

MartinSchuhmacher:
# Description

As a user I want to visit a random course and see the newly designed course overview page with the tabs "Lerninhalte", "Tools" and "Gruppen". The old tabs for "Themen" and "Aufgaben" will not be seen anymore and will be replaced by the view currently accessible through the room view button.

Ticket: https://ticketsystem.dbildungscloud.de/browse/BC-1393
Nuxt client PR: https://github.com/hpi-schul-cloud/nuxt-client/pull/2131
Server PR: https://github.com/hpi-schul-cloud/schulcloud-server/pull/3412
Dev branch PR: https://github.com/hpi-schul-cloud/dof_app_deploy/pull/228

## Approval for review


Metauriel (approved):

controllers/coursegroups.js
```diff
@@ -155,6 +155,8 @@ router.get('/:courseGroupId/', (req, res, next) => {
 		_.each(courseGroup.userIds, u => u.displayName = `${u.firstName} ${u.lastName}`);
 		_.each(course.teacherIds, t => t.displayName = `${t.firstName} ${t.lastName}`);
 
+		const showRoomView = Configuration.get('ROOM_VIEW_ENABLED') || false;
```
virgilchiriac, line 158: this should be added to the config json, otherwise is always false
https://github.com/hpi-schul-cloud/schulcloud-client/blob/main/config/default.json

also, should test if overwrite the config from env var with something like ROOM_VIEW_ENABLED=false
in some case, if the type is string and not boolean, it might evaluate "false" as being true

virgilchiriac (changes requested):

controllers/coursegroups.js
Metauriel: configuration should be extensively tested in other places, especially the handling of the string "false". That is responsability of the configuration service itself.
This flag in particular already exists in the client, so it doesnt need to be added again.

apart from that the || false is redundant, since Configuration.get('ROOM_VIEW_ENABLED') has a default value (false), and thus will never be undefined.

virgilchiriac: ah, my bad. I checked in wrong file. You're right about the part of the testing, just was thinking to avoid potential general problem

virgilchiriac (approved):

PR #2851, opened 2022-05-05
filtering "SchulcloudNextcloud" from ltiTools

IgorCapCoder:
# Description
hides Nextcloud in tool create menu in courses
Development in server is in progress

https://ticketsystem.dbildungscloud.de/browse/N21-89

## Approval for review

More and detailed information on the *definition of done* can be found [on Confluence](https://docs.dbildungscloud.de/pages/viewpage.action?pageId=92831762)


controllers/tools.js
```diff
@@ -31,6 +31,7 @@ const addToolHandler = (req, res, next) => {
 
 	api(req).get('/ltiTools', { qs: { isTemplate: true } })
 		.then((tools) => {
+			const filterTools = tools.data.filter((tool) => tool.name !== 'SchulcloudNextcloud');
```
Metauriel, line 34: Is there a way to add this logic to the server request? (potentially in conjunction with a parameter or seperate route to control if the user gets all tools, or only those he can add himself...)

Are there any other tools this is already solved for? 

please at the very least define a constant array of tools to be filtered. That can later then be refactored to come from somewhere else, like the server.

Metauriel (changes requested):

controllers/tools.js
alweber-cap: this logic will be added to the server with the tickets N21-90 and N21-91 when we do the associated refactoring tasks. Until then it is safer and less intrusive to filter in the frontend

alweber-cap (approved):

Metauriel (approved):

controllers/tools.js
Metauriel: 👍

PR #2884, opened 2022-05-27
BC-1824 - Fix blog card style

christian-darsow:
# Description
Article cards in the blog section on https://dbildungscloud.de/ should have the same width 

https://ticketsystem.dbildungscloud.de/browse/BC-1824

## Approval for review


static/styles/homepage/blog-card.scss
```diff
@@ -39,6 +40,7 @@ a.blog-card-wrapper{
     }
     @include media-breakpoint-down(md) {
         width: 100%;
+        overflow: visible;
```
MartinSchuhmacher, line 43: why is the overflow visible here?

christian-darsow: In general, we only need the overflow property as we mix flex layout with ellipsed text-overflow on a flex-item. As we change to one card per row rendering on smaller breakpoints (flex wrap), we need to make the overflow visible again in combination with width 100%. An alternative approach to get the expected rendering behaviour on smaller breakpoints is to set min-width of the flex-items to 100%. Well, it's all about the internal width rendering of the flex system

MartinSchuhmacher: Sounds reasonable

PR #2912, opened 2022-06-24
BC-1740 - Incorrect behaviour when creating a task with same start and submission date

MartinSchuhmacher:
# Description

When a Teacher creates a Task where the "visible for students and to be edited from" date is the same as the "processing period ends on" its seems like the creation was successful but its not.

The possibility to apply the same date for availableDate and dueDate on this page and saving should not be possible, the user should stay on this page.

BC-1740
## Approval for review


static/scripts/homework/edit.js
```diff
@@ -80,4 +71,16 @@ window.addEventListener('DOMContentLoaded', () => {
 			$('#teamsize').addClass('hidden-xl-down');
 		}
 	});
+
+	$('#homework-submit-btn').on('click', (event) => {
+		event.preventDefault();
+		const availableDate = $('#availableDate').val().split(' ');
+		const dueDate = $('#dueDate').val().split(' ');
+		if (dueDate[0] === '' || dueDate > availableDate) {
```
virgilchiriac, line 79: should check for length first? Also, the dueDate and availableDate are now arrays, not date. What do we compare?

MartinSchuhmacher: not sure about the length and if that is needed. but I saw that method to compare the different entries of an array somewhere else and I found that suitable for our case of the date. Not really sure how to do it in a different way

virgilchiriac (approved):

PR #2954, opened 2022-08-10
consent feature flag

odalys-dataport:
# Description

https://ticketsystem.dbildungscloud.de/browse/BC-2289

## Approval for review

### DIFF
--- a/helpers/theme.js
+++ b/helpers/theme.js
@@ -39,6 +39,7 @@ const setTheme = (res) => {
 	res.locals.theme = {
 		name: SC_THEME,
 		title: SC_TITLE,
+		short_title: SC_SHORT_TITLE,
 		theme_title: themeTitle,
 		institute_title: instituteTitle,
 		documents: {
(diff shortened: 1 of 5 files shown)


helpers/theme.js
```diff
@@ -30,7 +31,7 @@ const instanceInstitute = () => {
 };
 
 const instanceCloudConsentNecessary = () => {
-	if (SC_THEME === 'n21' || SC_THEME === 'brb') {
+	if (!Configuration.get('FEATURE_CONSENT_NECESSARY')) {
```
odalys-dataport, line 34: We can just return the value of the feature variable itself here, no?

virgilchiriac (approved):

helpers/theme.js
MartinSchuhmacher: I would agree to that, why not just return the configuration value?

christian-darsow: I shifted this change and also some variable renaming to the follow-up ticket BC-2276

MartinSchuhmacher (approved):

virgilchiriac (approved):

PR #2960, opened 2022-08-16
BC-2313 - Newly created class in the old school year Ticket

MartinSchuhmacher:
# Description

Currently the old school year is automatically selected during class creation. This is due to a wrong sorting in the first place and should be removed.

BC-2313
## Changes

* remove unnecessary sorting before showing available school years in class creation page
## Approval for review

### DIFF
--- a/controllers/administration.js
+++ b/controllers/administration.js
@@ -1538,8 +1538,7 @@ const renderClassEdit = (req, res, next) => {
 			const mode = req.locals.mode;
 			Promise.all(promises).then(
 				([teachers, gradeLevels, currentClass]) => {
-					let schoolyears = getSelectableYears(res.locals.currentSchoolData);
-					schoolyears = schoolyears.sort((a, b) => a.startDate.localeCompare(b.startDate));
+					const schoolyears = getSelectableYears(res.locals.currentSchoolData);
 
 					const allSchoolYears = res.locals.currentSchoolData.years.schoolYears
 						.sort((a, b) => b.startDate.localeCompare(a.startDate));


controllers/administration.js
```diff
@@ -1538,8 +1538,7 @@ const renderClassEdit = (req, res, next) => {
 			const mode = req.locals.mode;
 			Promise.all(promises).then(
 				([teachers, gradeLevels, currentClass]) => {
-					let schoolyears = getSelectableYears(res.locals.currentSchoolData);
-					schoolyears = schoolyears.sort((a, b) => a.startDate.localeCompare(b.startDate));
+					const schoolyears = getSelectableYears(res.locals.currentSchoolData);
```
christian-darsow, line 1541: There is probably no easy solution to have the years sorted asc && having the correct year pre-selected on creating a new class, or? This does the trick with correct preselection but the order of the options seems quiet weird, at least to me :)

MartinSchuhmacher: exactly, there would be a new logic needed to implement what we actually need. right now the array is filled with the "actual year" in first place and this we use in this solution to get what we need without puting too much effort into old logic

MartinSchuhmacher: so the question is: can we live with the order of the years or do we want to implement new logic to have a proper order there

christian-darsow: i would then agree, let us live with this order. Correct preselection should probably not make the user open the options anyway

christian-darsow (approved):

PR #3036, opened 2022-12-12
BC-1508 - Delete files on submission deletion

bischofmax:
# Description

## Approval for review

### DIFF
--- a/controllers/homework.js
+++ b/controllers/homework.js
@@ -398,7 +398,7 @@ router.delete('/:id/file', (req, res, next) => {
 router.get('/submit/:id/import', getImportHandler('submissions'));
 router.patch('/submit/:id', getUpdateHandler('submissions'));
 router.post('/submit', getCreateHandler('submissions'));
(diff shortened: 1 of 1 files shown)


controllers/homework.js
```diff
@@ -398,7 +398,7 @@ router.delete('/:id/file', (req, res, next) => {
 router.get('/submit/:id/import', getImportHandler('submissions'));
 router.patch('/submit/:id', getUpdateHandler('submissions'));
 router.post('/submit', getCreateHandler('submissions'));
-router.delete('/submit/:id', getDeleteHandler('submissions', true));
+router.delete('/submit/:id', getDeleteHandler('submissions'));
```
dyedwiper, line 401: I'm irritated that you removed the true here for two reasons:
1) The get route below is used by the delete link. As far as I see it, this delete route is unused can be removed entirely.
2) The redirect does not work without the true. So you must leave it in the get route.

PR #3070, opened 2023-01-13
Bc 1630 occasional csrf problems ipad

Michaellinaresxk:
# Description

## Approval for review


views/lib/csrf_content.hbs
```diff
@@ -1,8 +1,8 @@
 <div class="error-header-wrapper pb-3">
-    {{> 'lib/components/svg_error_content' }}
-    <h1 class="h2 error-header-code">CSRF</h1>
+    {{> 'lib/components/svg_error_to_much_tabs' }}
```
Metauriel, line 2: check spelling, should probably be "svg_error_too_many_tabs"

views/lib/csrf_content.hbs
```diff
@@ -1,8 +1,8 @@
...
+    {{!-- <h1 class="h2 error-header-code">CSRF</h1> --}}
```
Metauriel, line 3: if this line is no longer needed, feel free to delete it.
In case we want to bring it back we can always find it in the git history :P

locales/en.json
```diff
@@ -1611,7 +1611,8 @@
 			"submit": "Submit",
 			"toCourse": "To the course",
 			"tryAgain": "Try again",
-			"yes": "Yes"
+			"yes": "Yes",
+			"loginToSchulcloud": "Login to Schulcloud"
```
Metauriel, line 1615: please dont use "Schulcloud" but the name of the instance. Check in with UX or an experienced frontend developer on which variable to use, I dont know them by heart :)

locales/en.json
```diff
@@ -2285,7 +2286,10 @@
 				"unfortunatelyTheRequestFailedAgain": "Sorry, the request failed again. To continue using it, please reload the page.",
 				"weCouldNotMakeSureThat": "We could not make sure that this request really came from you. Please click the button below to reload the page.",
 				"weCouldNotMakeSureThatThisRequestReallyCameFromYou": "We could not make sure that this request really came from you. Please click the button below to confirm.",
-				"yourRequestCouldNotBeVerified": "Your request could not be verified."
+				"yourRequestCouldNotBeVerified": "Your request could not be verified.",
+				"yourSessionHasExpired": "Your session has expired",
+				"youAreAlreadyLoginInOtherTab": "It appears that you are already logged into Schulcloud in another tabs.",
```
Metauriel, line 2291: again use instance name instead of schulcloud

locales/en.json
```diff
@@ -1683,7 +1684,7 @@
 			"archive": "Archive",
 			"calendar": "Calendar",
 			"contact": "Contact",
-			"courseData": "My course files",
```
Metauriel, line 1686: why is this change here? Doesnt seem to be related to the ticket, and I am not sure its actually correct...

views/lib/csrf_content.hbs
```diff
@@ -1,8 +1,8 @@
...
     <p class="error-header-text">
-        {{$t "lib.csrf_content.text.yourRequestCouldNotBeVerified" }}
+        {{$t "lib.csrf_content.text.yourSessionHasExpired" }}
```
Metauriel, line 5: why not use the yourRequestCouldNotBeVerified anymore? The session is not technically expired (timed out), but denied for a different reason.

In case you really want to change the message here, please also remove the old text from the language files if its no longer used.

Metauriel (changes requested):
I really like your solution to the ticket, and its the correct approach to explain to the user what happens, and possible reasons why. However, it might still make sense to have the word CSRF somewhere...

locales/en.json
Michaellinaresxk: the course data still there i place something before then i remove the line maybe is just change the line but the data still there

locales/en.json
Michaellinaresxk: Ready

PR #3096, opened 2023-02-03
BC-3164-Added sort for all users

VicDavydiuk:
# Description
Added sorting function for all users for Teams

- https://ticketsystem.dbildungscloud.de/browse/BC-3164

## Approval for review

### DIFF
--- a/static/scripts/teamMembers.js
+++ b/static/scripts/teamMembers.js
@@ -8,13 +8,12 @@ $(document).ready(() => {
 		get: (searchParams, prop) => searchParams.get(prop),
 	});
 	const { sortDirection } = params;
-	const role = $('.dropdown-name').attr('current-user');
+	const thClass = '.is-sorted.table-wrapper .col-sort';
 
 	const sortByAscDesc = () => {
-		$('.col-sort').click((e) => {
+		$(thClass).click((e) => {
 			e.stopPropagation();
 			e.preventDefault();
-
 			const key = e.target.getAttribute('key');
 			const url = window.location.pathname;
 			if (key <= 2) {
@@ -29,8 +28,7 @@ $(document).ready(() => {
 
 	const className = sortDirection === 'desc' ? 'fa-caret-down' : 'fa-caret-up';
 
-	if (role === 'Administrator') {
-		$('.col-sort').each((index, elem) => {
+		$(thClass).each((index, elem) => {
 			if (index < 3) {
 				$(elem).append(
 					$(`<i class='col-sort-icon fa ${className}'></i>`),
@@ -39,7 +37,6 @@ $(document).ready(() => {
 		});
 
 		sortByAscDesc();
-	}
 
 	const handler = {
 		get(target, name) {
--- a/static/scripts/teamMembers.js
+++ b/static/scripts/teamMembers.js
@@ -17,10 +17,10 @@ $(document).ready(() => {
 			const key = e.target.getAttribute('key');
 			const url = window.location.pathname;
 			if (key <= 2) {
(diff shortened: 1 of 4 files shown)


views/lib/components/table.hbs
```diff
@@ -9,7 +9,7 @@
     </select>
     </div>
 {{/if}}
-<div class="table-wrapper {{#unless class}}table table-hover table-bordered{{/unless}}">
+<div class="{{#if isSorted}}is-sorted{{else}}{{/if}} table-wrapper {{#unless class}}table table-hover table-bordered{{/unless}}">
```
MateuszSapalaPassCon, line 12: "{{else}}" does nothing and can be removed

static/scripts/teamMembers.js
```diff
@@ -29,8 +28,7 @@ $(document).ready(() => {
 
 	const className = sortDirection === 'desc' ? 'fa-caret-down' : 'fa-caret-up';
 
-	if (role === 'Administrator') {
-		$('.col-sort').each((index, elem) => {
+		$(thClass).each((index, elem) => {
```
MateuszSapalaPassCon, line 31: Lint: Expected indentation of 1 tab but found 2.

MateuszSapalaPassCon (changes requested):

static/scripts/teamMembers.js
```diff
@@ -8,38 +8,35 @@ $(document).ready(() => {
...
 			e.preventDefault();
-
 			const key = e.target.getAttribute('key');
 			const url = window.location.pathname;
 			if (key <= 2) {
-				if (sortDirection === 'asc') {
-					window.location = `${url}?sortDirection=desc&sortBy=${key}`;
-				} else {
+				if (sortDirection === 'desc') {
 					window.location = `${url}?sortDirection=asc&sortBy=${key}`;
```
OliverHappe, line 21: if 'desc' -> sortDirection asc? is this intentionally?

VicDavydiuk: No, when we first went to the page, the default was ASC. And when we clicked on the sort button, it changed back to ASC. I changed it so that we didn't have to click on the button twice

OliverHappe: ok i understand that this solves the issue of setting the same value as the default as the initial click. does this fix solve the issue in a logical way?

i read this more as a hotfix that inverts the logic of sorting instead of fixing that misleading default. 

if the chosen value for sort is "desc" i would expect that the generated query-string is for "desc" sorting as well. As far as I understand this the chosen sort direction (after clicking the button for example) and the actual sort direction will be out of sync?

VicDavydiuk: No, it is synced.
For example:
1) We come to the page directly (5e1dba1eaa30ab4df47e11d2/members, without sorting params in url), it is sorted as ASC by default
2) Then we click on the th, and sort changes to DESC as far as url changes to 5e1dba1eaa30ab4df47e11d2/members?sortDirection=DESC&sortBy=0
3) Click again, and it changes to sort by ASC and url is 5e1dba1eaa30ab4df47e11d2/members?sortDirection=ASC&sortBy=0

The issue was with the empty url (empty means without sort parameters), so now it is fixed, and url param is fully synced with the grid (asc in url when grid sorted by asc, and desc in url when grid is sorted by desc)

OliverHappe: ok thanks for clarifying :)

OliverHappe (approved):

MateuszSapalaPassCon (approved):

PR #3101, opened 2023-02-15
BC-2419 Change unit of maximum file size to bytes

dyedwiper:
# Description

## Approval for review


static/scripts/dropzone/dropzone.js
```diff
@@ -119,7 +119,7 @@
       withCredentials: false,
       parallelUploads: 2,
       uploadMultiple: false,
-      maxFilesize: 256,
+      maxFilesize: 256000000,
```
SevenWaysDP, line 122: Why not 2684354560 ? Can you try, remove this prop?

dyedwiper: I didn't pay attention there. This is apparently just a default copied from Dropzone. I removed it and saw no difference locally. It shouldn't have any effect because before it was 256 Bytes.

SevenWaysDP (approved):

PR #3015, opened 2022-11-21
BC-1033 amp char not displayed properly in a course name field

bn-pass:
# Description

## Links to Tickets or other pull requests

https://ticketsystem.dbildungscloud.de/browse/BC-1033

## Changes

Add decoding of the course name retrieved from the BE as it was not properly displayed in an input field.

## Deployment

No new deployment requirements

## New Repos, NPM packages or vendor scripts

In the end there is no new dependency - at first I didn't notice that a very similar library **html-entities** (https://www.npmjs.com/package/html-entities) was already in the project. So then I and just replaced the **he** library added in the first commit  on this branch with **html-entities** as it does the same thing. So there is no new dependency from this bugfix after all.

## Screenshots of UI changes
image

No UI changes, just the behavioural change of adding proper decoding of the data displayed in an input form.

## Approval for review

- [x] QA: In addition to review, the code has been manually tested (if manual testing is possible)
- [x] All points were discussed with the ticket creator, support-team or product owner. The code upholds all quality guidelines from the PR-template.

> Notice: Please remove the WIP label if the PR is ready to review, otherwise nobody will review it.

PR #3291, opened 2023-08-29
BC-482-character-not-shown-properly

Michaellinaresxk:
Screen Shot 2023-08-29 at 2 17 54 PM
…reation.

# Description

https://ticketsystem.dbildungscloud.de/browse/BC-482

## Changes
Added functionality to display the character "&" instead of &amp on theme creation.

## Screenshots of UI changes
Edit

## Approval for review


bn-pass (changes requested):
I know the ticket's description only mentions the & character, but in my opinion the solution is too limited - please try to use some generic function to properly decode any possible special characters. Maybe you can use a similar [solution like for the BC-1033 ticket](https://github.com/hpi-schul-cloud/schulcloud-client/pull/3015/files) as the issue seems to be exactly the same.

bn-pass (changes requested):
It's better now, but please do this check-and-decode for each field separately - in this solution if, e.g., course description would be undefined, the course name would not be properly decoded. So just do it the same way, but for every field separately.

bn-pass (approved):

odalys-dataport (approved):

odalys-dataport (approved):

SevenWaysDP (approved):

PR #3313, opened 2023-09-22
BC-5119-Prevent create new neXboard (Front) 

Michaellinaresxk:
# Description
This is the fronted implementation to avoid teachers to create a new nexboard, removing the button of 
nextBoard, on edit-topic page.

https://ticketsystem.dbildungscloud.de/browse/BC-5119
https://github.com/hpi-schul-cloud/schulcloud-server/pull/4522

## Changes
- Remove the nextBoard button on edit-topic view for teacher role.
- Changed FEATURE_NEXBOARD_ENABLED to false

## Approval for review


views/topic/edit-topic.hbs
```diff
@@ -44,10 +44,9 @@
 
                 <div class="form-group">
                     <label>{{$t "topic._topic.label.content" }}</label>
-
                     <!-- React Magic -->
-                    <div id="content-blocks" data-value="{{lesson.contents}}" data-parent-id="{{lesson._id}}" data-school-id="{{schoolId}}" data-parent-type="lessons" data-etherpadbaseurl="{{etherpadBaseUrl}}"
-                        data-iscoursegroup="{{courseGroupId}}" data-nexboardenabled="{{#ifEnv "FEATURE_NEXBOARD_ENABLED" "true"}}true{{/ifEnv}}"></div>
+                    <div id="content-blocks" data-testid="nextboardButton" data-value="{{lesson.contents}}" data-parent-id="{{lesson._id}}" data-school-id="{{schoolId}}" data-parent-type="lessons" data-etherpadbaseurl="{{etherpadBaseUrl}}"
```
bn-pass, line 48: I'm not an expert here, but to me it looks like by adding `data-testid="nextboardButton"` you would mark the whole `div` as a nextboardButton (by the way it's Nexboard, not Nex**t**board) for our tests automation tool which is not right I think - it's a set of buttons, not a single button.

views/topic/edit-topic.hbs
```diff
@@ -44,10 +44,9 @@
...
+                        data-iscoursegroup="{{courseGroupId}}" data-nexboardenabled="{{#ifEnv "FEATURE_NEXBOARD_ENABLED" "false"}}true{{/ifEnv}}"></div>
```
bn-pass, line 49: Isn't this "true"/"false" a default value for the given env var? If that's correct, then this will not change anything for the instances that have the `FEATURE_NEXBOARD_ENABLED` enabled anyway (which should stay set to "true" there as we want to keep the Nexboards usable until we'll have another solution working).

bn-pass (changes requested):

config/global.js
```diff
@@ -30,6 +30,7 @@ const {
 	FEATURE_CONTACT_FORM_ATTACHMENTS_ENABLED,
 	FEATURE_MINT_PAGES_ENABLED,
 	FEATURE_NEXBOARD_ENABLED,
+	FEATURE_NEXBOARD_BUTTON_ENABLE,
```
bn-pass, line 33: Please use the past form, it should be ENABLE**D**.

config/global.js
```diff
@@ -73,6 +74,7 @@ const exp = {
...
```
bn-pass, line 77: Please use the past form, it should be ENABLE**D**.

views/topic/edit-topic.hbs
```diff
@@ -44,10 +44,9 @@
 
                 <div class="form-group">
                     <label>{{$t "topic._topic.label.content" }}</label>
-
                     <!-- React Magic -->
-                    <div id="content-blocks" data-value="{{lesson.contents}}" data-parent-id="{{lesson._id}}" data-school-id="{{schoolId}}" data-parent-type="lessons" data-etherpadbaseurl="{{etherpadBaseUrl}}"
-                        data-iscoursegroup="{{courseGroupId}}" data-nexboardenabled="{{#ifEnv "FEATURE_NEXBOARD_ENABLED" "true"}}true{{/ifEnv}}"></div>
+                    <div id="content-blocks" data-testid="nexboardButton" data-value="{{lesson.contents}}" data-parent-id="{{lesson._id}}" data-school-id="{{schoolId}}" data-parent-type="lessons" data-etherpadbaseurl="{{etherpadBaseUrl}}"
```
bn-pass, line 48: Why did you add the `data-testid` here? As I wrote in a comment before, I'm not an expert on this topic (should be verified by someone else), but doesn't it mean that the whole div will be treated as a **nexboardButton** and this is **not** a single button, but the container for all these buttons? @VikDavydiuk do you maybe know whether I'm right or not?

views/topic/edit-topic.hbs
```diff
@@ -44,10 +44,9 @@
...
+                        data-iscoursegroup="{{courseGroupId}}" data-neXboardButtonEnabled="{{#ifEnv "FEATURE_NEXBOARD_BUTTON_ENABLED" "false"}}true{{/ifEnv}}"></div>
```
bn-pass, line 49: You have a different env var name here and in the config file with the env vars - please remember to add the "D" at the end of the "FEATURE_NEXBOARD_BUTTON_ENABLE" in the app's env vars config.

static/scripts/topicEdit.js
```diff
@@ -326,7 +326,7 @@ class TopicBlockList extends React.Component {
      * Render the list items.
      */
 	render() {
-		const neXboardEnabled = ($contentBlocksContainer.data('nexboardenabled') === true);
+		const neXboardButtonEnabled = ($contentBlocksContainer.data('neXboardButtonEnabled') === false);
```
bn-pass, line 329: Shouldn't it be `=== true`? To me it looks that If the `neXboardButtonEnabled` will be `true` (in a case we do want to show the button) we will **not** show the button - it would have to be `false` which is not valid.

bn-pass (changes requested):

views/topic/edit-topic.hbs
bn-pass: @Michaellinaresxk why did you mark the conversation as resolved? Did you consult it with @VikDavydiuk or someone else? Please help me understand why did you add this `data-testid` param here.

views/topic/edit-topic.hbs
```diff
@@ -44,10 +44,9 @@
 
                 <div class="form-group">
                     <label>{{$t "topic._topic.label.content" }}</label>
-
                     <!-- React Magic -->
-                    <div id="content-blocks" data-value="{{lesson.contents}}" data-parent-id="{{lesson._id}}" data-school-id="{{schoolId}}" data-parent-type="lessons" data-etherpadbaseurl="{{etherpadBaseUrl}}"
-                        data-iscoursegroup="{{courseGroupId}}" data-nexboardenabled="{{#ifEnv "FEATURE_NEXBOARD_ENABLED" "true"}}true{{/ifEnv}}"></div>
+                    <div id="content-blocks" data-testid="nexboardButton" data-value="{{lesson.contents}}" data-parent-id="{{lesson._id}}" data-school-id="{{schoolId}}" data-parent-type="lessons" data-etherpadbaseurl="{{etherpadBaseUrl}}"
+                        data-iscoursegroup="{{courseGroupId}}" {{#unlessEnv "FEATURE_NEXBOARD_ENABLED" "true"}} data-neboardenabled="false" {{/unlessEnv}}></div></div>
```
bn-pass, line 49: I have a hard time understanding the logic here. Please correct me if I'm wrong, but to me it looks like that: unless the `FEATURE_NEXBOARD_ENABLED` will be set to `true`, the `data-neboardenabled="false"` (BTW you've got a type here, I guess it should be data-ne**x**boardenabled) will be added which basically disables the Nexboard button so in the end when `FEATURE_NEXBOARD_ENABLED` will be set to `false`, the `data-neboardenabled="false"` will not be added and thanks to this the button also will not be added as I guess it's only added when the `data-neboardenabled` is set to `true` (which also means it has to be present right)?

So does it mean the button will be always disabled, in any case so the clause doesn't really do anything?

config/http-headers.js
```diff
@@ -21,7 +21,7 @@ const config = {
 		// Default Content-Security-Policy Header for every site
 		// Use 'strict-dynamic' 'nonce-<nonceValue>' (nonceValue auto generated) to create a whitelist
 		corsDefault: {
-			defaultSrc: '\'self\' data: blob: wss://dbildungscloud.de wss://scchat.dbildungscloud.de https://api.dbildungscloud.de https://scchat.dbildungscloud.de https://s3.hidrive.strato.com https://libreoffice.dbildungscloud.de https://docs.dbildungscloud.de https://etherpad.dbildungscloud.de https://blog.niedersachsen.cloud https://blog.dbildungscloud.de https://sc-content-resources.hpi-schul-cloud.de https://open.hpi.de https://upload.wikimedia.org',
+			defaultSrc: '\'self\' data: blob: wss://dbildungscloud.de wss://scchat.dbildungscloud.de http://localhost:9001 https://api.dbildungscloud.de https://scchat.dbildungscloud.de https://s3.hidrive.strato.com https://libreoffice.dbildungscloud.de https://docs.dbildungscloud.de https://etherpad.dbildungscloud.de https://blog.niedersachsen.cloud https://blog.dbildungscloud.de https://sc-content-resources.hpi-schul-cloud.de https://open.hpi.de https://upload.wikimedia.org',
```
bn-pass, line 24: Why? I don't see a reason to add it here, please don't push anything that you had to do to just run something locally - it can create a vulnerability on PROD as CORS setting would allow the communication with user's localhost.

static/scripts/topicEdit.js
bn-pass, line 1: Please use `prettier` for every modified line of code (but not for every modified file as a whole). It looks like formatting got out of control here.

views/topic/edit-topic.hbs
```diff
@@ -62,7 +61,7 @@
                         {{closeLabel}}
                     </button>
 
-                    {{> "lib/components/csrfInput"}}
+                    {{"lib/components/csrfInput"}}
```
bn-pass, line 64: ?? Will it work without the `>` char after `{{`?

bn-pass (changes requested):

views/topic/edit-topic.hbs
Michaellinaresxk: The {{#unlessEnv "FEATURE_NEXBOARD_ENABLED" "true"}} ... {{/unlessEnv}} block dynamically adds data-**nexboardenabled="false"** to the <div> element only if the **FEATURE_NEXBOARD_ENABLED** is not set to true.
This means, if **FEATURE_NEXBOARD_ENABLED** is false or not defined in the environment, data-**nexboardenabled="false"** is added to the <div>. **If FEATURE_NEXBOARD_ENABLED** is true, this attribute won’t be added.

The JavaScript logic that renders the NeXboard button should check the data-nexboardenabled attribute. If it’s "false" or not present, the button won’t be rendered (or will be hidden), effectively controlling the button's visibility based on the FEATURE_NEXBOARD_ENABLED setting.

bn-pass: > If FEATURE_NEXBOARD_ENABLED is true, this attribute won’t be added.
Exactly. That's why the button will not be rendered. So it'll never be rendered, in any case.

static/scripts/topicEdit.js
bn-pass: Formatting (indentation to be exact) here still seems to be not correct.

views/topic/edit-topic.hbs
```diff
@@ -62,7 +61,7 @@
                         {{closeLabel}}
                     </button>
 
-                    {{> "lib/components/csrfInput"}}
+                    {{>"lib/components/csrfInput"}}
```
bn-pass, line 64:
```suggestion
                    {{> "lib/components/csrfInput"}}
```
Let's leave this clause untouched, please.

bn-pass (changes requested):

bn-pass (approved):

odalys-dataport (approved):

virgilchiriac (approved):

PR #3410, opened 2024-02-12
BC-6303 reduce startupProbe failureThreshold

Loki-Afro:
# Description

BC-6303

## Approval for review

### DIFF
--- a/ansible/roles/schulcloud-client-core/templates/deployment.yml.j2
+++ b/ansible/roles/schulcloud-client-core/templates/deployment.yml.j2
@@ -76,7 +76,7 @@ spec:
           exec:
             command: ["wget", "--timeout", "4", "-O", "/dev/null", "http://localhost:3100/login"]
           timeoutSeconds: 5
-          failureThreshold: 36
+          failureThreshold: 12
           periodSeconds: 5
         resources:
           limits:


bn-pass (approved):

virgilchiriac (approved):

ansible/roles/schulcloud-client-core/templates/deployment.yml.j2
```diff
@@ -76,7 +76,7 @@ spec:
           exec:
             command: ["wget", "--timeout", "4", "-O", "/dev/null", "http://localhost:3100/login"]
           timeoutSeconds: 5
-          failureThreshold: 36
+          failureThreshold: 12
           periodSeconds: 5
```
virgilchiriac, line 80: should probably introduce initialDelaySeconds

Loki-Afro: @virgilchiriac there is a reason for that.
having an `initialDelaySeconds` means that the pod won't become available as fast as if you would configure it to have one.
so decision here was regarding speed and availability

mamutmk5: We go an other way, we use an high failure threshold for the startup prob. 
The initialDelaySeconds is an fix ammount of seconds to wait.
We use the hieg failer threshold multiplie with the test period then is posible to get the pods faster up.
For exampel we wait 30 Seconds as initial delay seconds the the pod cann be first up after 30 seconds.
But if we use instand an Failer Threshold from 6 with 5 seconds Period so we cann have the pod after 5 second and latest after 30 seconds.
So if we have an deployment where only 25% of the pods restart a time and we have for 30 seconds initial dilay in best case all up after 120 seconds. On the other hand with the Failer threshold from 6 with an period of 5 seconds in the best case all is up after 20 seconds.

PR #3508, opened 2024-08-19
BC-7804 - prevent logging of headers

virgilchiriac:
# Description

BC-7804

## Approval for review

### DIFF
--- a/app.js
+++ b/app.js
@@ -250,7 +250,12 @@ app.use((err, req, res, next) => {
 		res.locals = {};
 	}
 	// prevent logging jwts and x-api-keys
-	delete error.options.headers;
+	if (error.options && error.options.headers) {
+		delete error.options.headers;
+	}
+	if (error.response && error.response.request && error.response.request.headers) {
+		delete error.response.request.headers;
+	}
 
 	if (Configuration.get('FEATURE_LOG_REQUEST') === true) {
 		const reqInfo = {
--- a/app.js
+++ b/app.js
@@ -247,9 +247,6 @@ const errorHandler = (err) => {
 	if (error.options && error.options.headers) {
 		delete error.options.headers;
 	}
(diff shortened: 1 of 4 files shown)


app.js
```diff
@@ -238,19 +238,28 @@ const isTimeoutError = (err) => err && err.message && (
...
 	error.statusCode = status;
 
-	if (!error.options) {
-		error.options = {};
+	// prevent logging jwts and x-api-keys
+	if (error.options && error.options.headers) {
+		delete error.options.headers;
+	}
+	if (error.response && error.response.request && error.response.request.headers) {
+		delete error.response.request.headers;
```
MartinSchuhmacher, line 251: anything in `error.options.headers` or `error.response.request.headers` that might become deleted but shouldnt be?

funny actually....why is it `error.response.request.headers` instead of `error.request.headers` ?

app.js
```diff
@@ -303,4 +312,10 @@ app.use((err, req, res, next) => {
 	});
 });
 
+process.on('unhandledRejection', (err, res, req, next) => {
```
EzzatOmar, line 315: Nice. It would be good to add 
`process.on('uncaughtException', (err, origin) => {
`

In both cases it is advised to close requested resources while handling the exception. Otherwise there is a chance that for example the HTTP request remains open.

https://stackoverflow.com/questions/5999373/how-do-i-prevent-node-js-from-crashing-try-catch-doesnt-work#comment9952914_5999381

EzzatOmar (dismissed):

app.js
virgilchiriac: true, this is unncecery / has no effect because the promise reject exception is handled by the logger outside express, in the process function
will be removed

MartinSchuhmacher (approved):
looks good to me

PR #3554, opened 2025-01-13
BC-8683 - Wrong plural for news and the time specification

MartinSchuhmacher:
# Description

Fixing wrong German language times in legacy code (e.g. on dashboard for news and tasks).

BC-8683
## Screenshots of UI changes

grafik

grafik

## Approval for review


Loki-Afro (changes requested):
can you verify that it is broken in the same way in a recent momentjs version?

MartinSchuhmacher:
> can you verify that it is broken in the same way in a recent momentjs version?
> 
> like the opposite, instead of maintaining the customization, get rid of them as moment js supports a lot of locals https://github.com/moment/moment/tree/18aba135ab927ffe7f868ee09276979bed6993a6/locale

I agree, but sadly no, `2.30.1` does not fix that for us. But I do not guarantee, that some of our weird legacy FE and scripting issues there breaks it in the end. But removing more then what I already did just breaks something else there, so I won't go forward with that alternative and in the end adjusted our customization. Thats how I proceed and ended up with that PR :)

Loki-Afro (approved):

PR #3556, opened 2025-01-14
remove display inline

f-g-s:


MartinSchuhmacher (approved):
with the additional mentions in chat regarding formating the changes are fine for me :)

PR #3592, opened 2025-02-12
BC-8974 make session cookie httponly and set samesite

Loki-Afro:
# Description

session, like req.session gets only accessed on server-side so the cookie is safe to set to httponly

https://ticketsystem.dbildungscloud.de/browse/BC-8974

## Approval for review


app.js
```diff
@@ -140,7 +140,9 @@ if (redisUrl) {
 const SIX_HOURS = 1000 * 60 * 60 * 6;
 app.use(session({
 	cookie: {
-		// TODO ...cookieDefaults,
+		httpOnly: true,
+		sameSite: Configuration.get('SESSION_COOKIE_SAME_SITE'),
+		secure: 'auto',
```
virgilchiriac, line 145: I think this should be a boolean

Loki-Afro: @virgilchiriac true documentation states you can set it to 'auto'
https://github.com/expressjs/session?tab=readme-ov-file#cookiesecure

> The cookie.secure option can also be set to the special value 'auto' to have this setting automatically match the determined security of the connection

config/default.schema.json
```diff
@@ -279,6 +279,11 @@
 			"default": "",
 			"description": "A string with concatenated user types (student, employee) that don't need to explicitly state their consent with the terms and conditions. \"employee\" means teachers and admins"
 		},
+		"SESSION_COOKIE_SAME_SITE": {
+			"type": "string",
+			"default": "strict",
+			"description": "express-session cookie samesite attribute"
+		},
```
virgilchiriac, line 286: I would put this in the COOKIE group just below

app.js
```diff
@@ -140,7 +140,9 @@ if (redisUrl) {
 const SIX_HOURS = 1000 * 60 * 60 * 6;
 app.use(session({
 	cookie: {
-		// TODO ...cookieDefaults,
+		httpOnly: true,
```
virgilchiriac, line 143:
```suggestion
		httpOnly: Configuration.get('COOKIE__ HTTP_ONLY),
```

app.js
virgilchiriac, line 145:
```suggestion
		secure: Configuration.get('COOKIE__SECURE'),
```

config/default.schema.json
Loki-Afro: you can not apply the same logic to all cookies and some need more protection than others

app.js
Loki-Afro: no, this should not even be configurable, having it set to false is broken from the beginning

virgilchiriac (approved):

virgilchiriac (approved):

PR #3831, opened 2026-06-03
BC-11598 use distroless image where possible

Loki-Afro:
# Description

https://github.com/hpi-schul-cloud/schulcloud-server/pull/6270
https://github.com/hpi-schul-cloud/tldraw-server/pull/107
https://github.com/hpi-schul-cloud/file-storage/pull/234
https://github.com/hpi-schul-cloud/h5p-server/pull/17

## Approval for review

- [ ] DEV: Every new component is implemented having accessibility in mind (e.g. aria-label, role property)


Dockerfile
```diff
@@ -1,57 +1,68 @@
...
     python3 \
     autoconf \
     automake \
-    build-base \
+    build-essential \
     nasm \
-    tzdata \
-    zlib-dev
+    zlib1g-dev \
+    && rm -rf /var/lib/apt/lists/*
```
dyedwiper, line 13: Why this?

Dockerfile
```diff
@@ -1,57 +1,68 @@
...
-COPY helpers /home/node/app/helpers
-COPY controllers /home/node/app/controllers
-COPY config /home/node/app/config
-COPY bin /home/node/app/bin
-COPY --from=git /app/version /home/node/app/static/version
-# "build" .. this basically throws out non relevant files for the theme under build and does scss to css stuff
-RUN export NODE_OPTIONS=--openssl-legacy-provider && node node_modules/gulp/bin/gulp.js clear-cache && node node_modules/gulp/bin/gulp.js
-
-CMD ["npm", "start"]
+# Remove devDependencies and build-only files to keep the production image small
```
dyedwiper: If the comment is only meant to explain npm prune, I think it is unnecessary. Or what is meant by "build-only files"?

Dockerfile
```diff
@@ -1,57 +1,68 @@
...
+ENV NODE_ENV=production
+ENV NO_COLOR="true"
+ENV SC_THEME=default
+
+# Copy the cleaned /app directory from builder
+COPY --from=builder /app ./
+
+USER nonroot
+
+EXPOSE 3100
```
dyedwiper, line 66: Why is this now necessary and wasn't before?

Dockerfile
Loki-Afro: delete apt cache, basic operation normal behavior, would be bad practice if not

Dockerfile
Loki-Afro: it is not necessary, but a hint to your container engine

Dockerfile
Loki-Afro: i'll keep it here simply as a fact that 99% of all node devs have no clue what the difference of a dev dependency and a non dev are. Additionaly the image would build fine without the prune line as it adds additional understanding of the intention here

Dockerfile
Loki-Afro: https://docs.docker.com/build/building/best-practices/#apt-get

dyedwiper: Thanks for the link! But isn't it actually irrelevant here because it is not in the final stage, so that it doesn't end up in the image anyway?
https://docs.docker.com/build/building/multi-stage/

Dockerfile
dyedwiper: I don't quite get that. To me the comment sounds as if `npm prune` would remove "build-only files" besides dev deps (whatever that would be), but it only removes dev deps. I think "Remove devDependencies to keep the production image small" would be precise and that would also be okay for me.

Dockerfile
Loki-Afro: and still the code gets copy pasted, or shagged around or would end up as warning anyways in some static code analyzer, it simply does not hurt

Dockerfile
Loki-Afro: refined comment

dyedwiper (approved):

dyedwiper (approved):

PR #3878, opened 2026-08-13
BC-12128 - blocking contact form

hoeppner-dataport:
# Description

BC-12128
https://github.com/hpi-schul-cloud/schulcloud-server/pull/6433

## Changes
- The validation of email-addresses was slightly improved in frontend-code at the only thing that differed from recommended pattern.
- only added test-cases to backend e-mail-adress-validation to illustrate the former problem

## Approval for review

### DIFF
--- a/static/scripts/base.js
+++ b/static/scripts/base.js
@@ -1,5 +1,5 @@
 // eslint-disable-next-line max-len
-const EMAIL_VALIDATION_PATTERN = "^[a-zA-Z0-9.!#$%&'*+\\/=?\\^_`\\{\\|\\}~\\-]+@[a-zA-Z0-9](?:[a-zA-Z0-9\\-]{0,61}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9](?:[a-zA-Z0-9\\-]{0,61}[a-zA-Z0-9])?)+$"
+const EMAIL_VALIDATION_PATTERN = "^[\\p{L}\\p{N}\\p{M}.!#$%&'*+\\/=?\\^_`\\{\\|\\}~\\-]+@[\\p{L}\\p{N}](?:[\\p{L}\\p{N}\\p{M}\\-]{0,61}[\\p{L}\\p{N}\\p{M}])?(?:\\.[\\p{L}\\p{N}](?:[\\p{L}\\p{N}\\p{M}\\-]{0,61}[\\p{L}\\p{N}\\p{M}])?)+$"
 
 /**
  * HELPER - addEventListener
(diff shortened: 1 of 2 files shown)


sebmue-dataport (approved):

static/scripts/base.js
```diff
@@ -1,3 +1,6 @@
+// eslint-disable-next-line max-len
+const EMAIL_VALIDATION_PATTERN = "^[a-zA-Z0-9.!#$%&'*+\\/=?\\^_`\\{\\|\\}~\\-]+@[a-zA-Z0-9](?:[a-zA-Z0-9\\-]{0,61}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9](?:[a-zA-Z0-9\\-]{0,61}[a-zA-Z0-9])?)+$"
```
virgilchiriac: no need to have double backslashes, except for escaping the .
but actually, write it as a Literal, is cleaner with single escape backslash for the .

```suggestion
const EMAIL_VALIDATION_PATTERN = /^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)+$/;
```

test/mocha/email.test.js
```diff
@@ -0,0 +1,117 @@
...
+	const invalidEmails = [
+		'',
+		'just-text',
+		'@example.org',
+		'me@',
+		'me@@example.org',
+		'me example@example.org',
+		'me@example..org',
+		'me@example.org ',
+		' me@example.org',
```
virgilchiriac, line 60: maybe we should simply trim the email for empty strings instead

test/mocha/email.test.js
```diff
@@ -0,0 +1,117 @@
...
+		'me@-example.org',
+		'me@example-.org',
+		'me@example_.org',
+		'me@example',
+		'x@a',
+		'peter@example',
+		`toolonglabel@${domainLabel64}.org`,
+		'user@[127.0.0.1]',
+		'user@.example.org',
+		'user@example.org.',
```
virgilchiriac, line 70: some  with hidden characters - not rendered in the browser 

```suggestion
		'user@example.org.',
		'test@domain.d​e, // "test@domain.d\u200Be"
```

static/scripts/base.js
hoeppner-dataport: As it is a property of an input field, i can not use a RegExp-Object here - but have to use a properly escaped string.

test/mocha/email.test.js
hoeppner-dataport: It already gets trimmed in the process.

this test only ensures the regex is correct - aside from the process.
It should only allows valid emails.

virgilchiriac (approved):

PR #3904, opened 2026-09-21
SVSDEV-149 - Adjust static file handling

cdarsowdataport:
# Description

- Shift static file handling to build process
- Removal of unmaintained dependency
- By that fix not working cache-invalidation of locales files

https://fwu-de.atlassian.net/browse/SVSDEV-149

## Approval for review

- [ ] DEV: Every new component is implemented having accessibility in mind (e.g. aria-label, role property)


middleware/assets.js
```diff
@@ -1,69 +1,88 @@
...
-		if (Configuration.get('FEATURE_ASSET_CACHING_ENABLED') === true) {
-			lazyInitialization();
-			return staticifyInstance.middleware(req, res, next);
-		}
-		return next();
-	});
+	app.use(
+		express.static(buildThemeAssetDir, {
+			setHeaders: (res, filePath) => {
+				if (Configuration.get('FEATURE_ASSET_CACHING_ENABLED') !== true) {
```
sebmue-dataport: Maybe we can already check for `ASSET_CACHING_MAX_AGE_SECONDS` here. Otherwise further down everything will be `no-cache` anyway, right?

helpers/handlebars/helpers/index.js
```diff
@@ -159,6 +159,7 @@ const helpers = () => ({
 	},
 	userIds: (users) => (users || []).map((user) => user._id).join(','),
 	getAssetPath: (assetPath) => getStaticAssetPath(assetPath),
+	getLocaleScriptPath: (lng) => getLocaleScriptPath(lng),
```
sebmue-dataport: Maybe `language` or `locale` since other variables are not abbreviated either. Or do we even need these functions themselves and a simple `getLocaleScriptPath,` would be sufficient?

middleware/assets.js
```diff
@@ -1,69 +1,88 @@
...
 function themeName() {
 	return Configuration.get('SC_THEME') || 'default';
 }
 
-let staticifyInstance = null;
-const localesDir = path.join(__dirname, '../locales');
 const buildThemeAssetDir = path.join(__dirname, `../build/${themeName()}`);
+const ASSET_MANIFEST_FILE = 'asset-manifest.json';
+let assetManifest = null;
+let hashedPaths = null;
```
sebmue-dataport: Would move these further up just below the imports/requires

middleware/assets.js
cdarsowdataport: Not sure if I get you right. First we check if caching is enabled at all. In the lines further down we first check if that file we want to serve is a hashed file at all. If not and some client explicitly asks for some un-hashed file we serve the "original" file with no-cache. so if you run a build, there exists always the two version in the build folder... Just thought, that there might be a use-case for that - maybe wrong :)

helpers/handlebars/helpers/index.js
cdarsowdataport: good point, changed that to `language` - and would like to stay with the function as it is now, so that the usage in the main template (default.hbs) stays easy

middleware/assets.js
cdarsowdataport: Adjusted to better readibility as you suggested. Thanks!!

hoeppner-dataport (approved):

